Issue #19, opened 2013-12-01
Stdlib 'wrapping' doesn't pass positional args to the stdlib logger


gcbirzan:
This is used (by us, anyway), extensively, because we just like pain. But, even so, it should be supported.

PR #23, opened 2013-12-16
#19 Pass positional arguments down to stdlib

gcbirzan:
Apparently, there's an error checking merge status :)


hynek:
Closed via #34.

Issue #2, opened 2013-09-17
Feature: journal output option


philips:
This might be a nice thing to add- a JournalLogger class:

https://github.com/systemd/python-systemd

Cool project! I wish I used more Python so I could make use of it.

cjw296:
I bet you do by now ;-)

hynek:
Almost! :)  We’re in the process of upgrading to Xenial and logging is a part of that.  I hope to get around this ASAP.

cjw296:
@hynek - glad to see you're alive, tried to say hi on irc but tumbleweed...

hynek:
heh just came back from PyCon ZA

paravoid:
[Daiquiri's JournalHandler code](https://github.com/jd/daiquiri/blob/080a7b9076ede4d4c5e300be13f4c9a34c271bb5/daiquiri/handlers.py#L60) is a good example of how that could be done. It's 38 lines of code, so it's not that complicated.

Essentially it boils down to just `journal.send(message, {"KEY1": "value1", "KEY2": "value2"}`, but with the quirk that it needs uppercase keys, and that there are some special well-known keys (like `PRIORITY`) that would be good to support.

paravoid:
Also, the systemd module provides a [logging.handler JournalHandler](https://github.com/systemd/python-systemd/blob/a402d08da41413507a4ebebf18de897f7ddf1dd2/systemd/journal.py#L504) that is reasonable. So one can rely on the logging module for this.

This example is gross, but works:
```python
    logging.basicConfig(format="%(message)s", handlers=(systemd.journal.JournalHandler(),))
    structlog.configure(
        processors=[
            structlog.stdlib.add_log_level,
            lambda _, __, d: {k.upper() if k != "event" else k: v for k, v in d.items()},
            structlog.stdlib.render_to_log_kwargs,
        ],
        context_class=structlog.threadlocal.wrap_dict(dict),
        logger_factory=structlog.stdlib.LoggerFactory(),
        wrapper_class=structlog.stdlib.BoundLogger,
    )

hynek:
I wonder if we should add this to http://www.structlog.org/en/stable/logging-best-practices.html?

hynek:
At this point I think this should be an external package. I don't care about this, it adds a dependency, and the headaches I have from standard library logging are bad enough.

I'll be happy to help anyone, who would like to tackle that.

andrewshadura:
To be fair, it can be implemented internally, without an external dependency.

[...]

PR #29, opened 2014-11-06
Issue 19 fix: Pass #2: Pass positional arguments down to stdlib

lakshmi-kannan:
Ref: https://github.com/hynek/structlog/pull/23

Apologies for opening another PR for same issue. I didn't want to pollute the original author's PR. 

I addressed all of your comments that I agreed with and added a commit. I retained the original author's commit as well. Tox tests pass on my box. 

I disagreed with one comment about adding pass through methods in stdlib BoundLogger (https://github.com/hynek/structlog/pull/23/files#diff-519ce9a9803f8404219d01de0afb0c1fR65). There are more than 5 such methods that are in stdlib logger. I was reading the code for the first time and instantly got what getattr was trying to do. I don't feel like it's magic. If you still insist on explicit pass through methods, I can add a commit. Please let me know.

### DIFF
--- a/structlog/stdlib.py
+++ b/structlog/stdlib.py
@@ -96,15 +96,11 @@ class BoundLogger(BoundLoggerBase):
 
     fatal = critical
 
-    def log(self, level, msg, *args, **kw):
-        self._proxy_to_logger(level, msg, *args, **kw)
-
-    def _proxy_to_logger(self, method_name, event=None, *event_args,
+    def _proxy_to_logger(self, method_name, event, *event_args,
                          **event_kw):
         if event_args:
             event_kw['positional_args'] = event_args
-        return super(BoundLogger, self)._proxy_to_logger(method_name, event,
-                                                         *event_args,
+        return super(BoundLogger, self)._proxy_to_logger(method_name, event=event,
                                                          **event_kw)
 
     def exception(self, event=None, **kw):
--- a/structlog/stdlib.py
+++ b/structlog/stdlib.py
@@ -62,37 +62,48 @@ class BoundLogger(BoundLoggerBase):
         )
 
     """
-    def debug(self, event=None, **kw):
+    def __getattr__(self, item):
+        return getattr(self._logger, item)
+
+    def debug(self, event=None, *args, **kw):
         """
         Process event and call ``Logger.debug()`` with the result.
         """
-        return self._proxy_to_logger('debug', event, **kw)
+        return self._proxy_to_logger('debug', event, *args, **kw)
 
-    def info(self, event=None, **kw):
+    def info(self, event=None, *args, **kw):
         """
         Process event and call ``Logger.info()`` with the result.
         """
-        return self._proxy_to_logger('info', event, **kw)
+        return self._proxy_to_logger('info', event, *args, **kw)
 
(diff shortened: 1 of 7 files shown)


wbolster:
Ok, @lakshmi-kannan are you planning to continue working on this?

Minor remark re. the performance gains with concrete methods vs magic `__getattr__()`: this only makes sense for the real logging methods; not for setup code like `addHandler()`.

structlog/stdlib.py
```diff
@@ -149,6 +198,31 @@ def __call__(self, *args):
...
+
+    """
+    def __init__(self, strip_positional_args=False):
+        self.strip_positional_args = strip_positional_args
+
+    def __call__(self, _, __, event_dict):
+        args = event_dict.get('positional_args')
+        if args:
+            args = tuple(args)
+            if len(args) == 1 and isinstance(args[0], dict) and args[0]:
```
wbolster: This code implements the same formatting logic as the stdlib's `logging` module does, which is basically applying the `%` operator, which supports both an iterable (tuple) and mapping (dict) argument.

The tuple conversion is needed in the current version of the code because the `%` operator is picky when it comes to lists:

```python
>>> '%d %d %s' % [1, 2, 'test']
  File "<stdin>", line 1, in <module>
TypeError: %d format: a number is required, not list
>>> '%d %d %s' % (1, 2, 'test')
'1 2 test'
```

But actually `tuple()` conversion above is not needed at all. `positional_args` is always a tuple since it comes from an `*args` function argument. However, the tests for this formatter provide a list instead, so with those updated the conversion can be removed.

wbolster:
Okay I have (hopefully) addressed all the comments from @hynek brought up in this PR in my own adaptation of it. I can't update this PR since the lakshmi-kannan:issue-19-fix branch is not mine, so I cannot push to it.

The new PR is #34.

structlog/stdlib.py
hynek: So IOW, this fragment was only there because the tests called the method wrong?  I figured something like that.

wbolster: yes.

Hynek Schlawack <notifications@github.com> schreef op 15 januari 2015 09:02:29 CET:
>> +    in the `positional_args` key. This is populated by
>> +    `structlog.stdlib.BoundLogger` or can be manually set.
>> +
>> +    `positional_args` can be any iterable, but a dictionary as the
>single
>> +    element of the tuple is used instead of the tuple, to mantain
>compatibility
>> +    with the undocumented feature of stdlib logging.
>> +
>> +    """
>> +    def __init__(self, strip_positional_args=False):
>> +        self.strip_positional_args = strip_positional_args
>> +
>> +    def __call__(self, _, __, event_dict):
>> +        args = event_dict.get('positional_args')
>> +        if args:
>> +            args = tuple(args)
>> +            if len(args) == 1 and isinstance(args[0], dict) and
>args[0]:
>
>So IOW, this fragment was only there because the tests called the
>method wrong?  I figured something like that.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/hynek/structlog/pull/29/files#r22994968

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

hynek:
Closed via #34.

PR #34, opened 2015-01-14
Positional args for stdlib logging

wbolster:
This PR addresses #19. It's a rebased version of PR#29 by @lakshmi-kannan on
top of current master, with additional commits by me that hopefully address all
the points raised by @hynek in the review of #29.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -24,6 +24,7 @@ The following folks helped forming ``structlog`` into what it is now:
 - `Tarek Ziadé <https://github.com/tarekziade>`_
 - `Thomas Heinrichsdobler <https://github.com/dertyp>`_
 - `Tom Lazar <https://github.com/tomster>`_
+- `Lakshmi Kannan <https://github.com/lakshmi-kannan>`_
 
 Some of them disapprove of the addition of thread local context data. :)
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -25,6 +25,7 @@ The following folks helped forming ``structlog`` into what it is now:
 - `Tarek Ziadé <https://github.com/tarekziade>`_
 - `Thomas Heinrichsdobler <https://github.com/dertyp>`_
 - `Tom Lazar <https://github.com/tomster>`_
+- `Wouter Bolsterlee <https://github.com/wbolster>`_
 
 Some of them disapprove of the addition of thread local context data. :)
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -19,12 +19,12 @@ The following folks helped forming ``structlog`` into what it is now:
 - `Itamar Turner-Trauring <https://github.com/itamarst>`_
 - `Jack Pearkes <https://github.com/pearkes>`_
 - `Jean-Paul Calderone <http://as.ynchrono.us>`_
+- `Lakshmi Kannan <https://github.com/lakshmi-kannan>`_
 - `Lynn Root <https://github.com/econchick>`_
 - `Noah Kantrowitz <https://github.com/coderanger>`_
 - `Tarek Ziadé <https://github.com/tarekziade>`_
 - `Thomas Heinrichsdobler <https://github.com/dertyp>`_
 - `Tom Lazar <https://github.com/tomster>`_
-- `Lakshmi Kannan <https://github.com/lakshmi-kannan>`_
 
 Some of them disapprove of the addition of thread local context data. :)
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,7 @@ Changelog
 - :release:`0.4.2 <2014-07-26>`
 - :bug:`8` Fixed a memory leak in greenlet code that emulates thread locals.
   It shouldn't matter in practice unless you use multiple wrapped dicts within one program that is rather unlikely.
+- :bug:`19` Pass positional arguments to stdlib wrapped loggers that use string formatting.
 - :feature:`-` :class:`structlog.PrintLogger` now is thread-safe.
 - :feature:`-` Test Twisted-related code on Python 3 (with some caveats).
 - :feature:`-` Drop support for Python 3.2.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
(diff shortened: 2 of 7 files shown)


wbolster:
The build seems to fail, but the only tox env that fails is the `docs` env:

```
Warning, treated as error:
<autodoc>:None: WARNING: more than one target found for cross-reference u'BoundLogger': structlog.BoundLogger, structlog.stdlib.BoundLogger, structlog.twisted.BoundLogger
```

The current master branch also shows this error, so this is unrelated to the changes in this PR.

wbolster:
With this branch, the following script:

```python
import logging

import structlog

logging.basicConfig(
    level=logging.DEBUG,
    format='%(asctime)s %(message)s')

structlog.configure(
    processors=[
        structlog.stdlib.filter_by_level,
        structlog.stdlib.PositionalArgumentsFormatter(
            strip_positional_args=True),
    ],
    context_class=dict,
)

logger.info("Normal message")
logger.info("Positional placeholder %d", 1)
logger.info("Positional placeholders %d %d", 1, 2)
logger.info("Keyword placeholders %(foo)d %(bar)s", dict(foo=1, bar='baz'))
```

...produces this output:

```
2015-01-14 16:44:42,091 {"event": "Normal message"}
2015-01-14 16:44:42,092 {"event": "Positional placeholder 1"}
2015-01-14 16:44:42,092 {"event": "Positional placeholders 1 2"}
2015-01-14 16:44:42,092 {"event": "Keyword placeholders 1 baz"}
```

hynek:
Great, thanks!  Will look into it first thing tomorrow!

hynek:
Meanwhile, @gcbirzan does this do what you need? :)

hynek:
I have fixed the build (tox installs with `--pre` and the new sphinx is weird…), please rebase.

AUTHORS.rst
```diff
@@ -19,6 +19,7 @@ The following folks helped forming ``structlog`` into what it is now:
 - `Itamar Turner-Trauring <https://github.com/itamarst>`_
 - `Jack Pearkes <https://github.com/pearkes>`_
 - `Jean-Paul Calderone <http://as.ynchrono.us>`_
+- `Lakshmi Kannan <https://github.com/lakshmi-kannan>`_
```
hynek: Please add yourself too.

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+- :feature`19` Pass positional arguments to stdlib wrapped loggers that use string formatting.
```
hynek: There’s a colon missing between `:feature` and the the first backtick.

structlog/stdlib.py
```diff
@@ -141,6 +231,35 @@ def __call__(self, *args):
...
+class PositionalArgumentsFormatter(object):
+    """
+    Applies stdlib-like string formatting to the `event` key with the arguments
+    in the `positional_args` key. This is populated by
+    `structlog.stdlib.BoundLogger` or can be manually set.
+
+    `positional_args` can be any iterable, but a dictionary as the single
+    element of the tuple is used instead of the tuple, to mantain compatibility
+    with the undocumented feature of stdlib logging.
+
```
hynek: No empty line here please (the pep rule comes from the fact that Guido’s editor needed it to reflow comments m().

structlog/stdlib.py
```diff
@@ -141,6 +231,35 @@ def __call__(self, *args):
         return logging.getLogger(name)
 
 
+class PositionalArgumentsFormatter(object):
+    """
+    Applies stdlib-like string formatting to the `event` key with the arguments
+    in the `positional_args` key. This is populated by
+    `structlog.stdlib.BoundLogger` or can be manually set.
+
+    `positional_args` can be any iterable, but a dictionary as the single
```
hynek: I think this `any iterable` isn’t true anymore, now we removed the call to `tuple()`? I’m also not sure what the rest of the comment means, do you? :confused:

tests/test_stdlib.py
```diff
@@ -137,6 +138,51 @@ def return_method_name(_, method_name, __):
         bl = BoundLogger(ReturnLogger(), [return_method_name], {})
         assert method_name == getattr(bl, method_name)('event')
 
+    def test_positional_args_proxied(self):
+        """
+        Positional arguments supplied must be proxied as kwarg.
```
hynek: …are proxied…

tests/test_stdlib.py
```diff
@@ -137,6 +138,51 @@ def return_method_name(_, method_name, __):
...
+        """
+        bl = BoundLogger(ReturnLogger(), [], {})
+        args, kwargs = bl.debug('event', 'foo', bar='baz')
+        assert kwargs.get('bar') == 'baz'
```
hynek: Could you switch this to `expected == actual` like the rest?

hynek:
Only minor stuff left except for the docstring of `PositionalArgumentsFormatter` which we need to understand/fix :)

Thank you so much to move this forward!

wbolster:
Okay, rebased onto current master, and added a few commits on top that should address all the comments @hynek made during the review above.

wbolster:
Btw, I think the `strip_positional_args` flag should be `True` by default. The whole point of positional args is that you can do string interpolation in your log messages; keeping the individual args in addition to the formatted message is nonsensical in most cases. The whole point of structlog is that messages have structure, so if you want to keep the individual pieces, string interpolation shouldn't be used as the means to achieve that.

wbolster:
@hynek fyi, the only non-trivial change is a8d32ec336b709ddea61e17664cc3c201b03c8bd

hynek:
I agree with the strip_positional_args argument, could you change that?

While at that a last nit: could you remove the `The` from “The `positional_args`”?

Thanks!

wbolster:
@hynek see 6532353 for the inverted behaviour; changed the name of the arg and names and comments in tests accordingly.

hynek:
Awesome, thank you!

I’ll cook up a release tomorrow!

hynek:
So I’ve just noticed we forgot to add posargs to `exception()`; does ca8c714965d37c3cad4e8de1e39e5bb9e4128fb0 look okay to you or did I miss some stdlib nuance?

wbolster:
lgtm

PR #37, opened 2015-01-16
README.rst: Wrap long lines

msabramo:
so everything is visible on a typical laptop display.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,10 +15,17 @@ structlog: Structured Logging for Python
 .. begin
 
 ``structlog`` makes structured logging in Python easy by *augmenting* your *existing* logger.
-It allows you to split your log entries up into key/value pairs and build them incrementally without annoying boilerplate code.
+It allows you to split your log entries up into key/value pairs and build them
+incrementally without annoying boilerplate code.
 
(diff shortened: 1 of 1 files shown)


hynek:
I’m sorry but that one is on purpose; it’s semantic linefeeds: http://rhodesmill.org/brandon/2012/one-sentence-per-line/  The correct way is using softwrap I guess?  I’m just coping with it in vim too.

msabramo:
Interesting. Haven't seen that before. So maybe it should be broken in a more semantic way -- is it okay to break lines at an `and` or at a comma for examples?

I think I like this:

```
It's dual-licensed under `Apache License, version 2 <http://choosealicense.com/licenses/apache/>`_
and `MIT <http://choosealicense.com/licenses/mit/>`_,
available from `PyPI <https://pypi.python.org/pypi/structlog/>`_,
the source code can be found on `GitHub <https://github.com/hynek/structlog>`_,
the documentation at `http://www.structlog.org/ <http://www.structlog.org>`_.
```

What do you think?

msabramo:
Could also be made shorter by externalizing the link URLs. So something like this:

```
It's dual-licensed under `Apache License, version 2`_ and `MIT`_,
available from `PyPI`_,
the source code can be found on `GitHub`_,
the documentation at `http://www.structlog.org/`_.

.. _Apache License, version 2: http://choosealicense.com/licenses/apache/
.. _MIT: http://choosealicense.com/licenses/mit/
.. _PyPI: https://pypi.python.org/pypi/structlog/
.. _GitHub: https://github.com/hynek/structlog
.. _http://www.structlog.org/: http://www.structlog.org/
```

PR #38, opened 2015-01-16
README.rst: Include example

msabramo:
and move `.. begin` so example doesn't appear twice in sphinx docs

Preview `README.rst` at https://github.com/msabramo/structlog/blob/README_teaser/README.rst

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,10 +12,19 @@ structlog: Structured Logging for Python
 .. image:: https://coveralls.io/repos/hynek/structlog/badge.png?branch=master
     :target: https://coveralls.io/r/hynek/structlog?branch=master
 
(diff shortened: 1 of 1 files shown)


hynek:
That removes the intro too though. *ponder*

I’d have to merge the pitch with the teaser I guess to make it work.

hynek:
Hm I tried it in https://github.com/hynek/structlog/tree/nicer-readme but nothing gets rendered.  I’ll retry with the next release in the hope that github has deployed the new lexer by then.

msabramo:
See https://github.com/hynek/structlog/pull/45

hynek:
Thanks. :)

PR #40, opened 2015-01-20
Improve suggested stdlib configuration (e.g. for structlog+logstash+forwarder)

wbolster:
This PR adds examples to the 'best practices' docs to get structlog to play nice with logstash and logstash-forwarder

### DIFF
--- a/docs/standard-library.rst
+++ b/docs/standard-library.rst
@@ -56,9 +56,8 @@ A basic configuration to output structured logs in JSON format looks like this::
         cache_logger_on_first_use=True,
     )
 
-If you plan to hook up the logging output to `logstash`, as suggested in :doc:`logging-best-practices`, you can simply output JSON, and have ``logstash-forwarder`` pick that up.
-To do so, you need to configure your process supervisor (such ``runit`` or ``supervisord``) to store the output in a file that is subsequently monitored by ``logstash-forwarder``, or alternatively you could pipe the output directly into ``logstash-forwarder``.
-To make your program behave like a proper `12 factor app`_ that outputs JSON to ``stdout``, configure the ``logging`` module like this::
+
+To make your program behave like a proper `12 factor app`_ that outputs only JSON to ``stdout``, configure the ``logging`` module like this::
 
     import logging
     import sys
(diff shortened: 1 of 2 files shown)


hynek:
Thanks!

Two things:

1. please don’t reflow the prose, I’m using intentionally [semantic linefeeds](http://rhodesmill.org/brandon/2012/one-sentence-per-line/).
2. Since the example is very heavy on stdlib and overlaps a lot with the example in standard-library.rst, I would suggest to integrate it there.

wbolster:
do you want me to provide an addition to the recommended stdlib configuration, and refer to it from the best practices page (which already mentioned logstash and logstash-forwarder)?

Hynek Schlawack <notifications@github.com> schreef op 21 januari 2015 06:52:04 CET:
>Thanks!
>
>Two things:
>
>1. please don’t reflow the prose, I’m using intentionally [semantic
>linefeeds](http://rhodesmill.org/brandon/2012/one-sentence-per-line/).
>2. Since the example is very heavy on stdlib and overlaps a lot with
>the example in standard-library.rst, I would suggest to integrate it
>there.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/hynek/structlog/pull/40#issuecomment-70788157

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

hynek:
yeah. something like “this is how you get stuff set up for logstash as mentioned in best practices” and best practices is a link

wbolster:
@hynek I've added the config snippets to the ``stdlib`` docs, and cross-referenced those from the best practices page. I've removed any duplication by just adding logstash-specific stuff below the suggested configuration. Though the examples mention logstash, I assume smart readers can easily adapt this to similar centralized logging solutions. :)  The main message is that outputting JSON (combined with a log shipper) is the simplest approach; e.g. much simpler than making ``logging`` talk over network sockets, and so on.

wbolster:
@hynek btw, regarding "semantic line feeds"... the future is here:

```
$ git diff --word-diff
```

;-)

hynek:
Hm, how about adding `add_log_level` and `add_logger_name` to the stdlib module?  Seems useful and saves a lot of words. :)

hynek:
oh and could you remove the whole “current incompatibilities” paragraph?  It’s not true anymore thanks to you. :)

wbolster:
@hynek i also noticed that and already opened #41 for that

wbolster:
i'm all for adding those two processors to the stdlib module, but i prefer to keep that separate from this docs PR.

i can take care of adding those processors properly, but i prefer not to have too many concurrent w-i-p PRs that need constant merging and rebasing, since it takes a lot of time and boring merging work.

wbolster:
oh, and the positional arguments processors should also be in the docs, and probably also in the suggested configuration pipeline. again; i prefer to do that after this docs-only PR goes in; it can be combined with the two other processors (which would also require docs)

hynek:
I completely agree on separating issues (although #41 have been okay :)), so two alternatives:

1. we/you add the processors to structlog.stdlib, rebase this PR and kill of the code for it,
2. we/you remove all mentions of the code, merge it and add it back later (together with the positional processor?) when it’s merged.

a) Which would you like better and b) would you like to do it?

wbolster:
i guess a) is the better approach. Let's leave this PR opne as an unfinished doc-only PR which is about logstash+forwarder config. This PR should be adapted once the suggested processors enter the stdlib.

I'll open separate tickets for adding the processors to the stdlib module.

wbolster:
See #43 for the follow-up that is about the processors (and not just about the doc changes suggested in this PR).

wbolster:
@hynek I've updated the doc changes to reflect the latest developments. The end result is much cleaner. This should cover everything discussed in #43.

hynek:
Looks good!

One more thing though: `add_*_*` ought to be added to the Processors section that currently claims there’s only one stdlib specific processor.  Could you fix that please?

docs/standard-library.rst
```diff
@@ -22,34 +22,63 @@ It behaves exactly like the generic :class:`structlog.BoundLogger` except:
...
+structlog comes with a few standard library-specific processors:
 
 :func:`~structlog.stdlib.filter_by_level`:
-   Checks the log entries's log level against the configuration of standard library's logging.
+   Checks the log entry's log level against the configuration of standard library's logging.
    Log entries below the threshold get silently dropped.
    Put it at the beginning of your processing chain to avoid expensive operations happen in the first place.
 
+:func:`~structlog.stdlib.add_logger_name`:
+   Adds the name of the logger to the output under the key ``logger``.
```
hynek: I guess I’m being pedantic here, but I think it should be “to the event dictionary” to be clear.

docs/standard-library.rst
```diff
@@ -22,34 +22,63 @@ It behaves exactly like the generic :class:`structlog.BoundLogger` except:
...
+
+:func:`~structlog.stdlib.add_log_level`:
+   Adds the log level to the output under the key ``level``.
```
hynek: same

hynek:
Thanks, I shall direct my wrath elsewhere. :)

wbolster:
Anyway, once this goes in you owe me a release ;)

hynek:
on it :)

PR #42, opened 2015-01-21
Add stdlib.BoundLogger.log

### LINKED ISSUE #36: BoundLogger for stdlib wrapper lacks .log() method
The summary says it all.

hynek:
Should fix #36?

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -76,7 +76,7 @@ structlog Package
 .. automodule:: structlog.stdlib
 
 .. autoclass:: BoundLogger
-   :members: bind, unbind, new, debug, info, warning, warn, error, critical, exception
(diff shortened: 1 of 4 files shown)


wbolster:
Without actually having tested it, this looks good and will handle the "standard" cases where the standard log levels are used. This probably covers most use cases, e.g. where the log level is dynamically determined and passed in as an argument to `.log(...)`.

However, the tricky part is when non-standard log levels are used. This is what I hinted at when I said I quickly gave up in #36. The stdlib supports defining custom log levels: https://docs.python.org/3/library/logging.html#levels

But perhaps structlog should just not support that. :)

hynek:
Well, if someone wants them, they’ll have to implement it. :)

Issue #43, opened 2015-01-21
Add more stdlib specific processors and recommend their use


wbolster:
Following discussion in #40, the processors that add the logger name and log level to the event should be added to structlog itself instead of living as mere suggestions in the docs. In addition the use of these processors should be recommended in the configuration example in the docs.  In addition to that, the recently added positional arguments processor from #19 and #34 should also make it into the suggested configuration.

What needs to be done is:

* [x] Move the `add_logger_name()` and `add_log_level()` processors from the suggested documentation changes in from #40 into the `structlog.stlib` module.
* [x] Add tests for those processors
* [ ] Update the recommended configuration in the docs to include these two processors, and  also the positional arguments formatter
* [ ] Rework the documentation changes in #40 so that it reflects the changes above

wbolster:
@hynek comments?

hynek:
sgtm

wbolster:
@hynek the tests can be pretty basic, right? e.g. just something trivial to assert that the key is added to the event_dict should be enough, right? it's not like they affect the whole pipeline.

hynek:
yeah just unit tests.  the nice thing about processors is that they are very simple to test.

wbolster:
Last two open points from the issue description should be covered in #40.

hynek:
yep!

PR #44, opened 2015-01-22
Add add_log_level() and add_logger_name() stdlib processors

wbolster:
See #43.

### DIFF
--- a/tests/test_stdlib.py
+++ b/tests/test_stdlib.py
@@ -21,6 +21,7 @@ from structlog.stdlib import (
     WARN,
     filter_by_level,
     add_log_level,
+    add_logger_name,
     _FixedFindCallerLogger,
 )
 from structlog._compat import PY2
@@ -259,3 +260,13 @@ class TestAddLogLevel(object):
         add_log_level(None, 'warn', event_dict)
         assert 'level' in event_dict
         assert 'warning' == event_dict['level']
+
+
+class TestAddLoggerName(object):
+    def test_logger_name_added(self):
+        name = 'sample-name'
+        logger = logging.getLogger(name)
+        event_dict = {}
+        add_logger_name(logger, None, event_dict)
+        assert 'logger' in event_dict
+        assert name == event_dict['logger']
--- a/tests/test_stdlib.py
+++ b/tests/test_stdlib.py
@@ -260,7 +260,6 @@ class TestAddLogLevel(object):
         The log level is added to the event dict.
         """
         event_dict = add_log_level(None, 'error', {})
-        assert 'level' in event_dict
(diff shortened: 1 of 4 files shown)


wbolster:
@hynek please review

hynek:
Thanks!

Couple very minor issues:

- add the new functions to api.rst (we forgot it in all other PRs :))
- …and to the changelog
- split the test of add_log_level into two please; one normal case and one alias. I try to test only one case per test.
- please test the return values of the functions, not the side-effect on the passed in `event_dict` (you don’t have to keep it around either)

Thanks again!

wbolster:
@hynek take 2

tests/test_stdlib.py
```diff
@@ -250,3 +252,33 @@ def test_positional_args_retained(self):
...
         assert positional_args == event_dict['positional_args']
+
+
+class TestAddLogLevel(object):
+    def test_log_level_added(self):
+        """
+        The log level is added to the event dict.
+        """
+        event_dict = add_log_level(None, 'error', {})
+        assert 'level' in event_dict
```
hynek: oops sorry i overlooked it in the first round: this is implicitly tested in the next line, let’s keep it simple

tests/test_stdlib.py
```diff
@@ -250,3 +252,33 @@ def test_positional_args_retained(self):
...
+        assert 'error' == event_dict['level']
+
+    def test_log_level_alias_normalized(self):
+        """
+        The normalized name of the log level is added to the event dict.
+        """
+        event_dict = add_log_level(None, 'warn', {})
+        assert 'level' in event_dict
```
hynek: same

tests/test_stdlib.py
```diff
@@ -250,3 +252,33 @@ def test_positional_args_retained(self):
...
+
+class TestAddLoggerName(object):
+    def test_logger_name_added(self):
+        """
+        The logger name is added to the event dict.
+        """
+        name = 'sample-name'
+        logger = logging.getLogger(name)
+        event_dict = add_logger_name(logger, None, {})
+        assert 'logger' in event_dict
```
hynek: same

hynek:
one more quick thing: I thing the changelog should refer the PR that actually got merged (i.e. 44) not a related issue.

Thanks again, will merge afterwards!

tests/test_stdlib.py
wbolster: I've removed it. Though it will fail with a `KeyError` instead of an `AssertionError` if the test fails. But I agree, simpler is better. (Hence my combined "alias" test, btw.)

hynek:
Thanks!

PR #45, opened 2015-01-24
README.rst: Change pycon to code-block

msabramo:
I think it wasn't rendering because the `pycon` is not a know reST directive.

With this hopefully, it will show up without highlighting and then with highlighting when GitHub pushes out their change. It should be highlighted on PyPI today if you register the new description.

Preview at https://github.com/msabramo/structlog/blob/patch-1/README.rst


hynek:
resolved per #38

PR #50, opened 2015-03-12
Add exception among supported levels in stdlib

leplatrem:
Using ``logger.exception(e)`` failed with processor ``structlog.stdlib.filter_by_level``


hynek:
Makes sense, thanks!

PR #51, opened 2015-03-13
Add option to specify key in TimeStamper processor

leplatrem:
In my environment, the logging standard is to store the human-readable timestamp in a ``'time'`` entry :)

### DIFF
--- a/structlog/processors.py
+++ b/structlog/processors.py
@@ -207,36 +207,37 @@ class TimeStamper(object):
         <http://en.wikipedia.org/wiki/ISO_8601>`_, or `None` for a `UNIX
         timestamp <http://en.wikipedia.org/wiki/Unix_time>`_.
     :param bool utc: Whether timestamp should be in UTC or local time.
+    :param str key: Property name stored in `event_dict`.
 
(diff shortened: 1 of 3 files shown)


structlog/processors.py
```diff
@@ -207,36 +207,37 @@ class TimeStamper(object):
         <http://en.wikipedia.org/wiki/ISO_8601>`_, or `None` for a `UNIX
         timestamp <http://en.wikipedia.org/wiki/Unix_time>`_.
     :param bool utc: Whether timestamp should be in UTC or local time.
+    :param str key: Property name stored in `event_dict`.
```
hynek: I don’t think ”Property” is the right term here.  Maybe something like “Target key in `event_dict` for added timestamps.”?

hynek:
Please:

- …fix the nit I pointed out above,
- …add a test doc string a la https://plus.google.com/+JonathanLange/posts/YA3ThKWhSAj
- …add a changelog entry.

Thanks!

hynek:
Thanks!  I was about to complain about `<unreleased>` but it actually works well.  I need to redo the changelog ASAP. :-/

PR #52, opened 2015-03-13
Support exception method in loggers

leplatrem:


hynek:
Thanks!

1. every new test requires a doc string like https://plus.google.com/+JonathanLange/posts/YA3ThKWhSAj
2. I think we should somehow make sure, that no method is missing.  How about using the keys of `_NAME_TO_LEVEL` sans `notset`?
3. Please use py.test’s parametrize instead of for loops: http://pytest.org/latest/parametrize.html  It makes nicer error reports.
4. I forgot about the changelog again.  Could you add an entry like “Better `log.exception()` within `structlog`.” and point it to both pull requests?

Thanks again!

leplatrem:
I'll update the PR, no problem.

Updating the changelog in the PR does not make it easier for the maintainer (and the contributor!), since it will often trivial conflicts.

IHMO updating it in the merge commit is a better approach :)

hynek:
speaking of which, could you rebase please :)

leplatrem:
Done!

hynek:
Thanks!

PR #58, opened 2015-09-15
Better support for missing __name__ in frame traversal

Callek:
In my own code, I got a new traceback for this issue (meant to be fixed in release of 15.1.0)

With some DEBUG prints sprinkled in, I was able to see that ``f.f_frame.has_key("__name__")`` was ``True``, however ``f.f_frame.get("__name__", "?")`` would still return ``None`` thus skipping that default choice.

Per python docs, ``foo.get("whatever")`` already has a default value of ``None`` so simply mixing them with "or" will achieve the desired goal here in all existing cases, as well as fix my own case.

For reference the full traceback I had:
```
Traceback (most recent call last):
  File "/data/www/relengapi/celery", line 81, in <module>
    load_entry_point('celery', 'console_scripts', 'celery')()
  File "/data/www/relengapi/virtualenv/lib/python2.7/site-packages/celery/__main__.py", line 30, in main
    main()
  File "/data/www/relengapi/virtualenv/lib/python2.7/site-packages/celery/bin/celery.py", line 81, in main
    cmd.execute_from_commandline(argv)
  File "/data/www/relengapi/virtualenv/lib/python2.7/site-packages/celery/bin/celery.py", line 769, in execute_from_commandline
    super(CeleryCommand, self).execute_from_commandline(argv)))
  File "/data/www/relengapi/virtualenv/lib/python2.7/site-packages/celery/bin/base.py", line 307, in execute_from_commandline
    return self.handle_argv(self.prog_name, argv[1:])
  File "/data/www/relengapi/virtualenv/lib/python2.7/site-packages/celery/bin/celery.py", line 761, in handle_argv
    return self.execute(command, argv)
  File "/data/www/relengapi/virtualenv/lib/python2.7/site-packages/celery/bin/celery.py", line 693, in execute
    ).run_from_argv(self.prog_name, argv[1:], command=argv[0])
  File "/data/www/relengapi/virtualenv/lib/python2.7/site-packages/celery/bin/worker.py", line 179, in run_from_argv
    return self(*args, **options)
  File "/data/www/relengapi/virtualenv/lib/python2.7/site-packages/celery/bin/base.py", line 270, in __call__
    ret = self.run(*args, **kwargs)
  File "/data/www/relengapi/virtualenv/lib/python2.7/site-packages/celery/bin/worker.py", line 212, in run
    state_db=self.node_format(state_db, hostname), **kwargs
[...]

PR #59, opened 2015-09-15
Fix assert-is-always-True in twisted tests

Callek:
This issue was found during my work on #58 (and is where I added myself to authors)

I don't feel this is worth adding to the changelog, and is the only place (other than #58) in which this was a problem afaict.

### DIFF
--- a/tests/test_twisted.py
+++ b/tests/test_twisted.py
@@ -101,12 +101,12 @@ class TestExtractStuffAndWhy(object):
         Extracts failures and events.
         """
         assert (
-            Failure(ValueError()), "foo", {} ==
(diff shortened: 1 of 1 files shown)


Callek:
...looks like this is failing after fixing the assert, I don't admit to knowing enough of twisted to get this test right, without simultaneously worrying that I'll also break the thing its trying to test though...

hynek:
Yeah test suite fixes don't need a change log entry.

The tests seem to fail now tho. :)

hynek:
I guess that `Failure`s aren't comparable.  Will look at it when I'm back.  Will probably need to split up the assert into an isinstance and comparing .value

PR #60, opened 2015-10-07
Adds JSON Formatter to Structlog's Stdlib Module

### LINKED ISSUE #9: Add a JSONifier for stdlib logging
Especially within frameworks, one hasn’t always the power to control every part of logging.

For Twisted, there is already a [solution in place](http://www.structlog.org/en/latest/twisted.html#bending-foreign-logging-to-your-will), there should be also one for stdlib.

fxfitz:
Sometimes when building your application you may need to use third party applications, and you cannot get them to point to `structlog.get_logger()`.  In which case, you may need to have your output as JSON.

So... bam!  I was wondering if I should be renaming `message` to `event` to better match how Structlog names things, so let me know about what you think.

Closes #9

### DIFF
--- a/docs/standard-library.rst
+++ b/docs/standard-library.rst
@@ -71,7 +71,10 @@ To make your program behave like a proper `12 factor app`_ that outputs only JSO
     import logging
     import sys
 
+    import structlog
+
     handler = logging.StreamHandler(sys.stdout)
+    handler.setFormatter(structlog.StdlibJSONFormatter())
     root_logger = logging.getLogger()
     root_logger.addHandler(handler)
 
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ CLASSIFIERS = [
     "Programming Language :: Python",
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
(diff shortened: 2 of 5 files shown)


hynek:
Ha so someone finally got around to do what I suggested. :)

So generally speaking, this looks too much of a hack (and adds a dependency) to be included in structlog proper.  But if you’d [put it on PyPI](https://hynek.me/articles/sharing-your-labor-of-love-pypi-quick-and-dirty/), I would happily link to it from the stdlib docs!

And yep, if someone is running this top have consistently JSON logs, I think it would make sense to either:

1. call it event
2. add an `event` K/V that makes it clear where the entry is coming from (something like `event='jsonified_external'` or whatever).

From the top of my head I would probably prefer option 2.

It also seems to me that attempting to `json.load` every single log entry is prohibitive performance-wise.

fxfitz:
Yeah, I wasn't sure how you'd feel about adding to your empty `INSTALL_REQUIRES` line, but I'd figure i'd go for it. ;-)

Anyway, I'll see if I can bake this in without having to depend on that package.

+1 for Option 2.

PR #67, opened 2016-04-25
drop missing keys from KeyValueRender

sibson:
Rather than render as None, simply drop any missing keys

### DIFF
--- a/src/structlog/processors.py
+++ b/src/structlog/processors.py
@@ -45,7 +45,7 @@ class KeyValueRenderer(object):
                 items = []
                 for key in key_order:
                     value = event_dict.pop(key, None)
-                    if value or not drop_missing:
+                    if value is not None or not drop_missing:
                         items.append((key, value))
                 items += sorted(event_dict.items())
                 return items
@@ -54,7 +54,7 @@ class KeyValueRenderer(object):
                 items = []
                 for key in key_order:
                     value = event_dict.pop(key, None)
-                    if value or not drop_missing:
+                    if value is not None or not drop_missing:
                         items.append((key, value))
                 items += event_dict.items()
                 return items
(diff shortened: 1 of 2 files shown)


hynek:
Hmmmm I feel like this should be a separate processor, such that other renderers benefit from it too.  Would you agree?

sibson:
Are you suggesting something like 

```
def drop_none_processor(event_dict):
    return dict(k: v for k, v in event_dict if v is not None)
```
If so the downside is that it would prevent logging None when its a valid value.  Otherwise, I don't understand your suggestion.

My particular problem is that KeyValueRenderer is injecting spurious new keys into my log messages when I set a key_order.  For example, web requests will have method,path keys which are never present in celery tasks but because Django shares the config I end up with method=None, path=None in my task logs.  I could use different config for each but that seems less DRY..

hynek:
Ah I see; sorry.  I thought you wanted to drop Nones altogether.

src/structlog/processors.py
```diff
@@ -31,27 +31,31 @@ class KeyValueRenderer(object):
...
-    def __init__(self, sort_keys=False, key_order=None):
+    def __init__(self, sort_keys=False, key_order=None, drop_missing=False):
         # Use an optimized version for each case.
         if key_order and sort_keys:
             def ordered_items(event_dict):
                 items = []
                 for key in key_order:
                     value = event_dict.pop(key, None)
-                    items.append((key, value))
+                    if value or not drop_missing:
```
hynek: This has to be `value is not None` or else `0` and `""` are dropped too.

src/structlog/processors.py
```diff
@@ -31,27 +31,31 @@ class KeyValueRenderer(object):
...
+                        items.append((key, value))
                 items += sorted(event_dict.items())
                 return items
         elif key_order:
             def ordered_items(event_dict):
                 items = []
                 for key in key_order:
                     value = event_dict.pop(key, None)
-                    items.append((key, value))
+                    if value or not drop_missing:
```
hynek: same

sibson:
@hynek updated to fix dropping "" and 0, thanks

hynek:
thank you!

PR #74, opened 2016-06-17
Configure PrintLogger to not append newlines

mattrobenolt:
As we were integrating with Sentry, we wanted to leverage our own msgpack renderer, but with msgpack, you wouldn't want to separate messages with newlines and this was causing problems.

So instead of copy/pasting the `PrintLogger` into our own, it'd be nice to just be able to provide this as a kwarg to the `PrintLoggerFactory`.

### DIFF
--- a/src/structlog/_loggers.py
+++ b/src/structlog/_loggers.py
@@ -21,16 +21,18 @@ class PrintLoggerFactory(object):
     To be used with :func:`structlog.configure`\ 's `logger_factory`.
 
     :param file file: File to print to. (default: stdout)
+    :param bool nl: Automatically append a newline to every message. (default: True)
 
     Positional arguments are silently ignored.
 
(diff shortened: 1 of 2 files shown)


hynek:
Just to go for sure: you really need to send the msgpack output to a file/stdout?  I’m asking because structlog allows for passing the event_dict structured into a logger which then can do whatever it wants.  There’s no need to make a detour via stdout.

Also you broke all builds. :)

hynek:
Um so the build is trivial to fix but did you decide on whether you really need it? :)  I may do a release around EuroPython-ish.

mattrobenolt:
wups, sorry for the lack of followup. Turns out, we ended up needing to use a different approach anyways.

hynek:
Cool. :)

PR #78, opened 2016-08-28
Add option to switch coloring output in ConsoleRenderer

insolite:
Could be useful for plain text file output.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,11 @@ The third digit is only for regressions.
 Changes:
 ^^^^^^^^
 
-*none*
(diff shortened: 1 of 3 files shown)


hynek:
So in principle, I always wanted ConsoleRenderer make the default (the default config and .dev [aren’t](http://www.structlog.org/en/stable/backward-compatibility.html) part of any backward compatibility promises), how about we make it default and if colorama is missing, it switches automatically to colorize=False (although I think we should it call just `colors`) and emits a warning?

insolite:
Yes, sounds good. I've commited all that things above.

hynek:
I took a slightly different route but thanks for the inspiration!

PR #80, opened 2016-09-14
Handling the case when exc_info is not True but evalues to True

IlianIliev:
According to the documentation [https://docs.python.org/2/library/logging.html#logging.Logger.debug] the exc_info argument should support not only boolean values, but also values evaluating to True.
This fix should solve the case when someone is using:

`log.debug('Debug message', exc_info=1)`

### DIFF
--- a/tests/test_processors.py
+++ b/tests/test_processors.py
@@ -447,25 +447,16 @@ class TestStackInfoRenderer(object):
 class TestFigureOutExcInfo(object):
     def test_obtains_exc_info_on_True(self):
         """
-        If True is passed, obtain exc_info ourselves.
-        """
-        try:
-            0/0
-        except Exception:
-            assert sys.exc_info() == _figure_out_exc_info(True)
-        else:
-            pytest.fail("Exception not raised.")
-
-    def test_obtains_exc_info_on_number(self):
-        """
         If the passed argument evaluates to True obtain exc_info ourselves.
         """
-        try:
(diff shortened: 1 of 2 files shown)


tests/test_processors.py
```diff
@@ -456,6 +456,17 @@ def test_obtains_exc_info_on_True(self):
         else:
             pytest.fail("Exception not raised.")
 
+    def test_obtains_exc_info_on_number(self):
```
hynek: I think it’s better to parametrize `test_obtains_exc_info_on_True` (and give it a better name) than to replicate almost everything.

IlianIliev: I have refactored the tests a bit. Currently it tests for True, int and float. I am not sure if there is a reason to test it with dicts, lists, objects etc. even the float looks like a pretty weird case to me.

hynek: your cases are fine but please use pytest’s parametrization for it – sorry I should have been more clear: http://docs.pytest.org/en/latest/parametrize.html (pro tip: they don't have to be tuples if it’s just one arg).

IlianIliev: No problem, I didn't know about that extra on pytest, maybe I should make myself more familiar with it. Thanks for the hint )))

hynek: Yep pytest is a treasure chest; spending some time in its docs pays off big time. :)

PR #89, opened 2016-11-21
Add note about third party processors to docs

underyx:
Feel free to request changes or even decline. I just thought it'd be easier to ask the question 'can/should this be added to the docs' in the form of a PR.

### DIFF
--- a/docs/processors.rst
+++ b/docs/processors.rst
@@ -116,3 +116,15 @@ Advanced log aggregation and analysis tools like `logstash <https://www.elastic.
 
 More examples can be found in the :ref:`examples <processors-examples>` chapter.
 For a list of shipped processors, check out the :ref:`API documentation <procs>`.
+
+
+Third Party Packages
+--------------------
+
+Since processors are self-contained callables,
+it's easy to write your own and to share them in separate packages.
+The following processor packages are known to be currently available on PyPI:
+
+- `structlog-pretty <https://github.com/underyx/structlog-pretty>`_: Processors for prettier output -- a code syntax highlighter, JSON and XML prettifiers, a multiline string printer, and a numeric value rounder.
+
+Please feel free to submit a pull request to extend this list with *your* package!


hynek:
I wonder if it weren’t better to activate the wiki for `structlog` for just this purpose so it’s smaller barrier for people to add their integrations. 🤔

underyx:
It _could_ make sense but I'm not sure if it matters that much and content in the wiki might be more difficult to find/reach.

I mean, package authors probably won't consider it a big deal to submit a PR editing the docs, and as it stands, it not as if there's that many people extending structlog anyway (sadly). The only reason I'd see for preferring the wiki is if you want to get rid of the maintainer overhead, in which case go for it!

hynek:
Heh so this PR violates quite a few of our style guidelines.   Would you like me to point them out to your for future reference or should I just fix it myself?  I’d feel like a dick to just start bickering on such a small change.

Tangentially, we should probably encourage people to add their own packages too so they don’t think you need some special blessing to do it?

underyx:
@hynek no worries, I'm available right now to fix the style issues. And I'm a maintainer of coala, where we complain until every commit is perfect, so at least I'll get to see the other side of this style of code review :D

So, fire away!

docs/processors.rst
```diff
@@ -116,3 +116,10 @@ Advanced log aggregation and analysis tools like `logstash <https://www.elastic.
 
 More examples can be found in the :ref:`examples <processors-examples>` chapter.
 For a list of shipped processors, check out the :ref:`API documentation <procs>`.
+
+Third Party Packages
```
hynek: Two blank lines before new sections.

docs/processors.rst
```diff
@@ -116,3 +116,10 @@ Advanced log aggregation and analysis tools like `logstash <https://www.elastic.
...
+--------------------
+
+Since processors are self-contained callables, it's easy to write your own and to share them in separate packages. The following processor packages are currently available on PyPI:
```
hynek: Semantic new lines.

docs/processors.rst
```diff
@@ -116,3 +116,10 @@ Advanced log aggregation and analysis tools like `logstash <https://www.elastic.
...
+
+- `structlog-pretty <https://github.com/underyx/structlog-pretty>`_: Processors for prettier output—a code syntax highlighter, JSON and XML prettifiers, a multiline string printer, and a numeric value rounder.
```
hynek: I can’t believe this will ever come up in a code review (but it makes me happy). :D  I’m firmly in the “space-endash-space” camp. :D  In RST it’s better to just write `output -- a` and it will be rendered correctly.  Props for Oxford comma though. :)

docs/processors.rst
hynek: also maybe let’s phrase it as  “are currently known to be available”

docs/processors.rst
hynek: as mentioned, there should be some kind of encouragement to add your own stuff after the enumeration.

hynek:
ah heh the only python package with a viral video :)

underyx:
Alright @hynek, thank you, I've actually learned a lot from this!

I made the changes to the best of my knowledge, if you're still unhappy with parts of the commit (the semantic newlines for instance), I'd like to ask to fix them yourself, as it'll probably be less work to explain with a diff than with a review 🙃

hynek:
Nah it’s fine.  I usually don’t do new lines after commas but that’s nothing I’d enforce.

underyx:
Just thought I'd mention that I've been using semantic newlines ever since and I just love them.

Thanks for introducing them to me, @hynek!

hynek:
Ha you're welcome.  I consider them indentation for prose: language structure is visually apparent. :)

PR #90, opened 2016-11-25
[Doc] Handle log formatting for third party components

gilbsgilbs:
We should be using `record.getMessage()` instead of `record.msg` to ensure that third party libraries logs are formatted.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -48,7 +48,7 @@ How can I make third party components log in JSON too?
               self._log = structlog.get_logger()
 
           def emit(self, record):
-              self._log.log(record.levelno, record.msg, name=record.name)
+              self._log.log(record.levelno, record.getMessage(), name=record.name)
 
       root_logger = logging.getLogger()
       root_logger.addHandler(StructlogHandler())


hynek:
Oh interesting!  Would you mind pointing me to some docs?  My knowledge of stdlib’s corner cases is mostly superficial.

gilbsgilbs:
Here is the official doc from logging module: https://docs.python.org/3/library/logging.html#logging.LogRecord.getMessage

The typical use-case is when third party app does something like this:
```
import logging
logger = logging.getLogger(__name__)
logger.info("foo %s", "bar")
```
This should output to `foo bar.`. With the current suggested implementation, it would output to `foo %s`. `getMessage()` is aimed at doing the formatting job. That's what this PR is about.

@multani in #9 also suggested to add exc_info logging right there. That's not the point of this PR, but I also reckon that this would be a good idea. And maybe integrate this directly inside structlog (don't know if it's already planned).

Thanks.

hynek:
Thanks!  Feel free to open more PRs to improve stdlib docs, but try to always include some background information. :)

PR #91, opened 2016-11-25
Flake8 fixes

underyx:
This might come in handy so that new PRs don't fail by default 🙃

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,7 @@ def read(*parts):
     with codecs.open(os.path.join(HERE, *parts), "rb", "utf-8") as f:
         return f.read()
 
+
 try:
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -17,7 +17,7 @@ commands = coverage run --parallel -m pytest {posargs}
 [testenv:flake8]
 basepython = python3.5
 skip_install = true
-deps = flake8
+deps = flake8~=3.2.0
```
hynek: Please don’t do this; it’s OK if stuff fails on releases. :)  Otherwise thanks, will merge once this is gone.

hynek:
thanks!

PR #92, opened 2016-11-27
Fix stdlib errors being logged as exceptions sometimes.

gilbsgilbs:
stdlib error log records were sometimes logged with exception level, sometimes
logged with error level by structlog. This is due to the non 1-1 mapping
between log levels and log names (some levels have the same name). If the
developer adds a new log name that conflicts with an already existing log
level without blacklisting it, it produces the unexpected behavior. It's even
random in Python 3: because of the hash seed, it will always produce the same
result in the same python interpreter instance, but will vary according to the
interpreter instance.

The problem is that once the mistake is done, it's pretty complicated to debug.
And it's even more complicated to write a test (since you will never notice a
behavior change in the same python instance). I think it would be a good idea
to add a disclaimer message at least. Feel free to add a test for this to
prevent future errors and time loss if you have any idea on how to write it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,9 @@ Changes:
 
 - Add *colors* argument to ``structlog.dev.ConsoleRenderer`` and make it the default renderer.
   `#78 <https://github.com/hynek/structlog/pull/78>`_
+- Fix bug with Python 3 and ``structlog.stdlib.log`` function. Error log level was not reproductible and was logged as exception
+  one time out of two.
+  `#92 <https://github.com/hynek/structlog/pull/92>`_
 
 
 ----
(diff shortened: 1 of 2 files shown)


gilbsgilbs:
Just to be 100% clear. With the previous implementation, `_LEVEL_TO_NAME` could either output to:

```
{
     ERROR: 'error',
    … other entries …
 }
```
(expected)
or to:

```
{
     ERROR: 'exception',
    … other entries …
 }
```
(unexpected)

This means that the `stdlib.BoundLogger` class could either log errors to error level (expected) or to exception level (unexpected). Having a quick look at the code, `stdlib.BoundLogger.log()` seems the only method being impacted by this bug.

You'll probably notice that this PR does not contain any test. It's not about lazyness, but this one is incredibely hard to write. The best test I could write would be a test that:
- Fails only one time out to two before the fix.
- Fails only on PY3 (even if the bug might exist on PY2) before the fix.
- Will not ensure that the mistake won't reproduce with another log level.

I don't write this kind of Schrœdinger

hynek:
Sorry I didn’t get around to look at this earlier.  It looks important and should have a changelog entry I think.

gilbsgilbs:
@hynek Done.

hynek:
thank you and sorry i didn’t get around to this earlier!

PR #93, opened 2016-11-27
Fix typo

gilbsgilbs:


hynek:
That's not a typo.  It means "if and only if". :)

gilbsgilbs:
Uuuh, for some reason, I did not expect that here at all :sleeping: . I probably forgot to activate my neurons. Sorry about that.

PR #95, opened 2016-12-03
Suggestion for using stdlib BoundLogger with third-party apps.

gilbsgilbs:
Hi there,

Here are some suggestions to improve stdlib. It allows to use the stdlib.BoundLogger, even when defining a root handler (to pass third-party apps logs through structlog's pipeline). Not sure my approach is really logical and elegant though (maybe closer to a POC) and suggestions/improvements are welcome. 

- Add a `stdlib.Handler` class that helps forward third-party apps logs through structlog. May fit to most use cases. However, can lead to recursion issues if you persist on using `stdlib.LogFactory`. To overcome this, I did some extra-work (see second bullet-point).

  Typical use-case: output third-party apps + structlog logs to stdout.
```python
import logging
import sys

import structlog


root_logger = logging.getLogger()
handler = structlog.stdlib.Handler()
root_logger.addHandler(handler)
root_logger.setLevel(logging.INFO)

structlog.configure(
    processors=[
        # structlog.processors.JSONRenderer(),
        structlog.dev.ConsoleRenderer(),
    ],
    context_class=dict,
)

logging.getLogger('foo').warning("logging!")
logging.getLogger('foo.bar').warning("logging again!")

structlog.get_logger().warning("we can also log from structlog!")
```

- Add a `stdlib._HandlerWrapper` class and a `stdlib.wrap_handler` function that allows to output structlog log through logging modules handlers (this required some more changes in `stdlib.LoggerFactory` to work properly). This wrapper prevents structlog from recursing.

Typical use-case: output third-party apps through structlog, and even output everything to vanilla handlers.

```python
import io
import logging
import sys

import structlog


root_logger = logging.getLogger()
s = io.StringIO()
handler = structlog.stdlib.wrap_handler(logging.StreamHandler(s))()
root_logger.addHandler(handler)
root_logger.setLevel(logging.INFO)

structlog.configure(
    processors=[
        # structlog.processors.JSONRenderer(),
        structlog.dev.ConsoleRenderer(),
    ],
    context_class=dict,
    logger_factory=structlog.stdlib.LoggerFactory(handlers=root_logger.handlers),
)

logging.getLogger('foo').warning("logging!")
logging.getLogger('foo.bar').warning("logging again!")

structlog.get_logger().warning("we can also log from structlog!")

print(s.getvalue())  # Everything was logged
```

Thus, this allows to use stdlib.filter_by_level and probably some other stuff I did not think of.

Cheers.

### DIFF
--- a/src/structlog/stdlib.py
+++ b/src/structlog/stdlib.py
@@ -358,9 +358,9 @@ class LoggerFactory(object):
         logger = logging.getLogger(name)
 
         if self._handlers is not None:
-            logger.propagate = False  # We don't propagate structlog logs to
-                                      # prevent them from being infinitely
-                                      # processed by a parent _WrappedLogger.
+            # We don't propagate structlog logs to prevent them from being
+            # infinitely processed by a parent _WrappedLogger.
+            logger.propagate = False
             self._override_handlers(logger)
 
         return logger
--- a/src/structlog/stdlib.py
+++ b/src/structlog/stdlib.py
@@ -14,6 +14,7 @@ from __future__ import absolute_import, division, print_function
 import logging
 
 from structlog._base import BoundLoggerBase
+from structlog._config import get_logger
 from structlog._frames import _find_first_app_frame_and_name, _format_stack
 from structlog.exceptions import DropEvent
 
@@ -103,8 +104,7 @@ class BoundLogger(BoundLoggerBase):
 
     fatal = critical
 
-    def _proxy_to_logger(self, method_name, event, *event_args,
-                         **event_kw):
+    def _proxy_to_logger(self, method_name, event, *event_args, **event_kw):
         """
         Propagate a method call to the wrapped logger.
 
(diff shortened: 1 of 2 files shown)


hynek:
Sorry I didn’t get around to look at this before.  Do I see it correctly, that it’s basically an attempt to codify the FAQ entry about getting stdlib logs into structlog?  It’s kind of funny that there’s an PR open to do vice versa at the same time. :)

gilbsgilbs:
@hynek 
>  Do I see it correctly, that it’s basically an attempt to codify the FAQ entry about getting stdlib logs into structlog?
Not only. This is an attempt to fix #9 . It allows more things than the doc suggests (e.g. output logs to any stdlib handler ; make possible to use stdlib filters such as `structlog.stdlib.filter_by_level`). It tries overcomes the infinite-recursion issue by not propagating the handler.

hynek:
How relevant would you see this with the new approaches that have been [introduced in 17.1](http://www.structlog.org/en/stable/standard-library.html#suggested-configurations)?  ISTM this is a closed chapter but I don’t want to just close the PR without consulting with you.

gilbsgilbs:
@hynek That is just great :) . Thanks! 👍

PR #101, opened 2017-02-01
Fix dev.ConsoleRenderer wrt. key 'event'

ma-ric:
handle the cases if 'event' is missing in event_dict or not a string

### DIFF
--- a/src/structlog/dev.py
+++ b/src/structlog/dev.py
@@ -168,7 +168,7 @@ class ConsoleRenderer(object):
                 # can be a number if timestamp is UNIXy
                 self._styles.timestamp + str(ts) + self._styles.reset + " "
             )
-        level = event_dict.pop("level",  None)
+        level = event_dict.pop("level", None)
         if level is not None:
             sio.write(
                 "[" + self._level_to_color[level] +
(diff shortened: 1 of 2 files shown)


ma-ric:
Locally I only run tests for py36, so I was not aware of the failing test related to py2 and utf string.

Is there a reason that the "event" is never repr'ed, as stated in the test, or could I change the test?

ma-ric:
wrapped "event" inside `str` instead of `self._repr`, so all tests passes

hynek:
Yeah the reason is that ConsoleRenderer is intended to be mainly human-readable and quotes around the event don’t look too nice.

What is your overall use case/intention?

ma-ric:
Yes this aspect I realized already. Therefore I placed `str` around "event". 

The intention to handle non-string events is only for convinces at writing debug messages at development stage.  An intermediate logging processor could make use of non-string "event". 

A use case for a missing "event" could be: ` log.debug(**some_dict)`

hynek:
So I thought about this (way too long admittedly :|) and I’m going to decline it.  ConsoleRenderer is intended to be human-readable and event is supposed to have special meaning.  Your use case can be simply achieved using `log.debug("dump", **somedict)`.

PR #112, opened 2017-05-07
Fix add_logger_name processor not working in foreign_prechain.

### LINKED ISSUE #111: add_logger_name processor does not work in ProcessorFormatter's pre_chain
Hi,

### Steps to reproduce
1. Take [this](http://structlog.readthedocs.io/en/stable/standard-library.html?highlight=add_logger_name#rendering-using-structlog-based-formatters-within-logging) example from the doc.
2. Add `structlog.stdlib.add_logger_name` processor to pre_chain.
3. Run the example.

### Expected behaviour
The logger name (`root` in the doc example) should be added to the event dict with the key `logger`.

### Actual behaviour
Crashes as soon as you log anything using logging module. Here is the stacktrace:
```
>>> logging.getLogger('foo').warning('bar')
--- Logging error ---
  File "/usr/lib64/python3.6/logging/__init__.py", line 992, in emit
    msg = self.format(record)logging.getLogger('foo').warning("bar")
  File "/usr/lib64/python3.6/logging/__init__.py", line 838, in format
    return fmt.format(record)
  File "/home/gilbsgilbs/Projects/structlog/src/structlog/stdlib.py", line 423, in format
    ed = proc(None, meth_name, ed)
  File "/home/gilbsgilbs/Projects/structlog/src/structlog/stdlib.py", line 349, in add_logger_name
    event_dict['logger'] = logger.name
AttributeError: 'NoneType' object has no attribute 'name'
Call stack:
  File "test.py", line 66, in <module>
    logging.getLogger('foo').warning('bar')
Message: 'bar'
Arguments: ()
```

### Comments
This is because `ProcessorFormatter` passes `None` as the logger to pre_chain processors. From inside the processor, we have lost track of the record, which was our only hope to get the logger name (`record.name`).

How this should be fixed according to you? I can contribute to fix this faster. Probably won't use `ProcessorFormatter` for now because of this issue.

gilbsgilbs:
Fixes #111

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,8 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-*none*
+- ``structlog.stdlib.add_logger_name()`` now works in ``structlog.stdlib.ProcessorFormatter``'s ``foreign_pre_chain``.
+  `#112 <https://github.com/hynek/structlog/issues/112>`_
 
 
 ----
--- a/src/structlog/stdlib.py
+++ b/src/structlog/stdlib.py
@@ -346,10 +346,11 @@ def add_logger_name(logger, method_name, event_dict):
     """
     Add the logger name to the event dict.
     """
-    if "_record" in event_dict:
-        event_dict["logger"] = event_dict["_record"].name
-    else:
+    record = event_dict.get("_record")
+    if record is None:
         event_dict["logger"] = logger.name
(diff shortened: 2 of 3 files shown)


hynek:
I wonder if we should promote the `_record` field as an official API because I’m sure there’s a lot useful things that could be done with it. *ponder*

hynek:
Thanks!

PR #114, opened 2017-05-07
Allow stdlib configuration to be serializable.

gilbsgilbs:
This merge request allows the stdlib configuration to be serializable, so that it can lay a bit more in configuration files instead of python code.

If it's convincing, I'd like to generalize this to all structlog configuration and even allow serializable processors definition, so that the configuration work can be in its right place, that is to say in configuration files. The final goal is to avoid using ugly conditionals to configure structlog. We should be able to just provide a self-explanatory configuration file.

### DIFF
--- a/docs/standard-library.rst
+++ b/docs/standard-library.rst
@@ -127,14 +127,20 @@ You can leave rendering for later and yet use ``structlog``’s renderers for it
             "disable_existing_loggers": False,
             "formatters": {
                 "plain": {
-                    "()": structlog.stdlib.ProcessorFormatter,
-                    "processor": structlog.dev.ConsoleRenderer(colors=False),
+                    "()": "structlog.stdlib.ProcessorFormatter",
+                    "processor": "structlog.dev.ConsoleRenderer",
+                    "processor_kwargs": {"colors": False},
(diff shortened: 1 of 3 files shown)


hynek:
I'm sorry I didn't comment on this for years but that's simply because I just didn't know what to do about it.

My overall opinion is that I don't like this kind of serialization for various reasons. However I can see that some people like it. So is there anything that could be done to achieve the same thing _outside_ of `structlog`?

IOW I would be willing to add stuff that would facilitate outside work, but not the work itself.

gilbsgilbs:
Hi @hynek. Thanks for the bump (and for your amazing talk at PyCon 2019 :heart: ).

We ended up writing a (private) third-party library that handles the hassle of configuring structlog with stdlib for us (mainly because we didn't want to copy-paste configuration related code across all of our applications). Even though I can't say I'm very proud of this solution, I don't think there's really anything preventing anyone from doing the specific job of this PR outside of structlog. It just feels a bit awkward but it works.

I don't mind if we close this. I probably should have closed it a long time ago.

hynek:
Cheers, and thanks!

PR #115, opened 2017-05-08
Support DropEvent in foreign_pre_chain

### LINKED ISSUE #113: DropEvent does not work in foreign_pre_chain
### Steps to reproduce
1. Take [this](http://structlog.readthedocs.io/en/stable/standard-library.html?highlight=add_logger_name#rendering-using-structlog-based-formatters-within-logging) example from the doc.
2. Append a processor that raises `DropEvent` in `pre_chain`.
3. Run the example.

### Expected behaviour
The log should be skipped.

### Actual behaviour
Raises DropEvent. The exception is not caught.

### Comments
I don't know if this is really expected or not. There should be at least a disclaimer in the documentation that states clearly that you cannot raise DropEvent in pre_chains processors.

However, I reckon there is a way to handle the DropEvent in pre_chain. We could declare a custom `Handler` that proxies everything to the `StreamHandler` (or whatever handler is configured), except that we catch/ignore the `DropEvent` exception in the overidden `emit` method.

gilbsgilbs:
Hi,

Here is a proof that it's possible to support `DropEvent` in `foreign_pre_chain` processors without too much overhead.  This fixes #113. It's based on my work on #114.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,3 +4,4 @@ freezegun>=0.2.8
 pretend
 pytest
 simplejson
+twisted
--- a/docs/standard-library.rst
+++ b/docs/standard-library.rst
@@ -127,14 +127,20 @@ You can leave rendering for later and yet use ``structlog``’s renderers for it
             "disable_existing_loggers": False,
             "formatters": {
                 "plain": {
-                    "()": structlog.stdlib.ProcessorFormatter,
-                    "processor": structlog.dev.ConsoleRenderer(colors=False),
+                    "()": "structlog.stdlib.ProcessorFormatter",
+                    "processor": "structlog.dev.ConsoleRenderer",
+                    "processor_kwargs": {"colors": False},
                     "foreign_pre_chain": pre_chain,
+                    # alternatively
+                    # "foreign_pre_chain": "mymodule.pre_chain",
                 },
                 "colored": {
-                    "()": structlog.stdlib.ProcessorFormatter,
-                    "processor": structlog.dev.ConsoleRenderer(colors=True),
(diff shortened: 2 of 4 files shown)


hynek:
I hope to be able to look at the serialization stuff soonish, but I don’t think it’s worth it to do this complex dance to allow for drop event in pre_chain.   We'll just document that it doesn’t work and be done with it.   Ppl who want to filter their stdlib entries can use stdlib’s mechanisms…

gilbsgilbs:
@hynek Thanks for your reply.

I thought it'd be less confusing for the user if everything worked in `pre_chain` just as it works for other processors, as long as possible. Correct me if I'm wrong, but I think it was kind of the original idea behind `pre_chain`.

Using logging module filters is indeed possible, but deports (and possibly duplicates) code into the user's code base (that's another dance). Additionally, it forces users to 1) figure out why it doesn't work (because users will never read the tiny bottom lines in the documentation) , 2) dive into stdlib filters stuff, when the problem could (and afaic should) be addressed by structlog itself because it provides a super-elegant manner to fix this  (way better than stdlib). In the end, I'll probably have to reproduce something similar to this MR into my app, just to avoid inheriting `logging.Filter` and find some unobvious ways to share my structlog/stdlib filters code. Moreover, stdlib filters might not apply easily in some complex cases.

About the Handler/Formatter dance🕺, you are right. It was one of my main concerns here. To avoid it, we could handle everything in the ProcessorHandler, which would deprecate the ProcessorFormatter. Then, you'd just define a wrapped handler and no formatters at all. No more dance.

I do think that `pre_chain` and `processors` should be able to merge in the end. That would be less confusing in a lot of respects.

It's up to you.
Thanks, 
Gilb's

PR #117, opened 2017-05-13
Clear args on LogRecords after rendering them in ProcessorFormatter

### LINKED ISSUE #116: Suggested configurations leads to exceptions for third party libs.
Hi,

Python 2.7.5
CentOS Linux release 7.2.1511 (Core)

I tried the suggestion named:

> Rendering Using structlog-based Formatters Within logging

From:
[http://www.structlog.org/en/stable/standard-library.html#rendering-using-structlog-based-formatters-within-logging](http://www.structlog.org/en/stable/standard-library.html#rendering-using-structlog-based-formatters-within-logging)

The only thing I added was:

```
import boto3  
ec2 = boto3.client('ec2')
response = ec2.describe_regions()
print('Regions:', response['Regions'])
response = ec2.describe_availability_zones()
print('Availability Zones:', response['AvailabilityZones'])
```
(This issue is not related to the third party lib used, I got the same thing with other third party libs, just needed one as a minimal repro case.)

This leads to a ton of exception that ends with:

> File "/usr/lib64/python2.7/logging/__init__.py", line 328, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting

My goal is to use structlog to get all my logs and those from imported third party libs to be formatted in JSON with a basic set of common keys. But I started with using the documentation as is first to understand how it works.

I was expecting boto3 logs to be formatted and it worked, somewhere in the console output (and the log file) there is:

> 2017-05-12 19:57:08 [info     ] Starting new HTTP connection (1): 169.254.169.254

What I don't understand are all the Exception thrown from inside the logging library that are in visible in the std err output. Should the suggested config work in this scenario?

Attached are the source code file used, the log file from the WatchedFileHandler and console output file where the exceptions are visible. It was generated with:
`python issue.py &> console_output.txt`

[console_output.txt](https://github.com/hynek/structlog/files/997889/console_output.txt)
[filehandler_output.txt](https://github.com/hynek/structlog/files/997891/filehandler_output.txt)
[source_code.txt](https://github.com/hynek/structlog/files/997890/source_code.txt)

Martin

hynek:
This fixes #116 and *probably* http://stackoverflow.com/questions/43855507/configuring-and-using-structlog-with-django .

@if-fi did I miss something obvious that will break everything for everyone? also cc @gilbsgilbs who seems to know a lot more about stdlib logging than I. :)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,7 @@ Changes:
 - Clear log record args in ``structlog.stdlib.ProcessorFormatter`` after rendering.
   This fix is for you if you tried to use it and got ``TypeError: not all arguments converted during string formatting`` exceptions.
   `#116 <https://github.com/hynek/structlog/issues/116>`_
+  `#117 <https://github.com/hynek/structlog/issues/117>`_
 
 
 ----
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,6 +26,9 @@ Changes:
 
 - ``structlog.stdlib.add_logger_name()`` now works in ``structlog.stdlib.ProcessorFormatter``'s ``foreign_pre_chain``.
   `#112 <https://github.com/hynek/structlog/issues/112>`_
+- Clear log record args in ``structlog.stdlib.ProcessorFormatter`` after rendering.
+  This fix is for you if you tried to use it and got ``TypeError: not all arguments converted during string formatting`` exceptions.
+  `#116 <https://github.com/hynek/structlog/issues/116>`_
 
 
 ----
--- a/src/structlog/stdlib.py
+++ b/src/structlog/stdlib.py
@@ -420,6 +420,7 @@ class ProcessorFormatter(logging.Formatter):
             logger = None
             meth_name = record.levelname.lower()
             ed = {"event": record.getMessage(), "_record": record}
+            record.args = ()
 
             # Non-structlog allows to run through a chain to prepare it for the
             # final processor (e.g. adding timestamps and log levels).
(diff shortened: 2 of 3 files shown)


if-fi:
Hi @hynek I just found the same problem on Thursday, and I was looking for a non-hackish solution to send as a pull request. 
Yours is pretty smooth. I just tried it locally and it works fine in all scenarios I came up with.

P.S. I also found a problem when using `ProcessorFormatter` with `RotatingFileHandler`. I will send you a pull request that I hope will be included in the same bugfix release as this one

gilbsgilbs:
Hi @if-fi, hi @hynek.

As far as I can see, this won't break anything and will fix the issue. Still, I have some remarks:

- You're clearing `record.args` too often. Clearing after calling `record.getMessage()` (which is the one actually consuming the args) should be sufficient I guess (untested),
- Not really fund of mutating `record.msg` and now `record.args`. You really shouldn't ever have to mutate any `record` attribute, appart from `record.message` (≠ from `record.msg`, yeah, ik, logging module is real crap, but `record.msg` should be the unformatted message and `record.message` the formatted one -_-"),
- You're actually formatting the message twice. First time in `record.getMessage()`, and another time calling super function, which puts the formatted message into `record.message` . The second time should be near-free since `record.args` is empty (provided that format operator is a bit smart, haven't checked that out). Yet, it still requires two format when one is most likely possible (calling super function sooner, and getting the formatted message in `record.message`).

Don't have much time to dig into this. I think the fix is ok.

hynek:
1. Good catch, I’ll move it up.
2. I sadly don’t know enough about this part.  But I guess I’d be hoping for too much if I thought I could format into `record.message` and it solves #119 too, right? :)
3. The reason is https://github.com/hynek/structlog/pull/105#discussion_r107939889

gilbsgilbs:
2. Not hoping too much at all. That's the right way to fix #119.

PR #118, opened 2017-05-13
Fix exc_info and race condition in ProcessorFormatter

hynek:
Based on / fixes #109 

Hey @andreiko I tried to push it to your branch but I failed at git hard (due to the changes in master) so here’s a new pull request with some changes (and split up of stack_info & exc_info handling).

Does this still fix your problem and do you agree with the approach?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,9 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
+- ``structlog.stdlib.ProcessorFormatter`` now accepts *keep_exc_info* and *keep_stack_info* arguments to control what to do with this information on log records.
+  Most likely you want them both to be ``False`` therefore it's the default.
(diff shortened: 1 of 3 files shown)


hynek:
(also @gilbsgilbs  – any opinions?  should be straight-forward, right?)

gilbsgilbs:
Just read it and seems fine and straightforward indeed. I did not test it because I'm away from keyboard (heading to Japan in a minute for 10 days).

I'm 99.99% sure it can't help at any rate in this case, but just to be sure, have a look at `handleError` method somewhere in `logging` (in `Handler` I think).

hynek:
Thank you, enjoy Japan!

andreiko:
This totally fixes my problem. Thank you for fixing the conflicts! Sorry that I wasn't available during the weekend.

hynek:
No problem at all, I’m glad we’ve sorted this out.  17.2 is on PyPI.

PR #120, opened 2017-05-14
Treat treat string log messages as dicts if they are in json format (#119)

### LINKED ISSUE #119: Using `ProcessorFormatter` with `RotatingFileHandler` misformats the events
I have the following configuration:

``` python

structlog.configure(
    processors=[
        structlog.processors.UnicodeEncoder(encoding='utf-8', errors='backslashreplace'),
        structlog.processors.TimeStamper(fmt='iso'),
    ],
    context_class=structlog.threadlocal.wrap_dict(dict),
)

pre_chain = [
    # Add the log level and a timestamp to the event_dict if the log entry
    # is not from structlog.
    structlog.processors.TimeStamper(fmt='iso'),
]

LOGGING_CONFIG = {
    'version': 1,
    'disable_existing_loggers': False,

    'formatters': {        
        'plain_struct': {
            "foreign_pre_chain": pre_chain,
            "processor": structlog.dev.ConsoleRenderer(colors=False, pad_event=0),
        },
        'json_struct': {
            "foreign_pre_chain": pre_chain,
            "processor": structlog.processors.JSONRenderer(sort_keys=True),
        }
     },

    'filters': {
    },

    'handlers': {
        # Handler to log all the Website generic information
        'log_generic': {
              'class': 'logging.handlers.RotatingFileHandler',
              'level': 'DEBUG',
              'formatter': 'plain_struct',
              'filters': [],
              'maxBytes': 1024*1024,
              'filename': os.path.join(app.config.get('LOG_LOCATION'), 'app_debug.log')
        },

        # Handler to log all the Website generic information in json format
        'log_json': {
            'class': 'logging.handlers.RotatingFileHandler',
            'level': 'DEBUG',
            'formatter': 'json_struct',
            'filters': [],
            'maxBytes': 1024 * 1024,
            'filename': os.path.join(app.config.get('LOG_LOCATION'), 'app_json.log')
        },
    },

    'loggers': {
        # logs website errors and debug logs
        'fbu': {
            'handlers': ['log_json', 'log_generic'],
            'level': 'DEBUG',
            'propagate': False
        },
    }
}
```

Due to some specifics in the `RotatingFileHandler` internal implementation, the log messages go through formatting twice, and I end up with those log messages in both files:

`app_debug.log`:
``` bash
2017-05-14T21:03:26.563866Z [info     ] 2017-05-14T21:03:26.563745Z [info     ] {"event": "{\"event\": \"I am here\", \"level\": \"info\", \"logger\": \"fbu.process\", \"timestamp\": \"2017-05-14T21:03:26.562886Z\"}", "level": "info", "timestamp": "2017-05-14T21:03:26.563333Z"}
```

`app_json.log`:
``` python
{"event": "{\"event\": \"I am here\", \"level\": \"info\", \"logger\": \"fbu.process\", \"timestamp\": \"2017-05-14T21:03:26.562886Z\"}", "level": "info", "timestamp": "2017-05-14T21:03:26.563333Z"}
```

if-fi:
(Fixes #119 )

Due to the nature of how RodatingFileHandler is implemented, the log recors are formatted twice.
Once in `def shouldRollover(self, record):` in `class RotatingFileHandler` and second time when `emit` is actually called after that
https://github.com/python/cpython/blob/master/Lib/logging/handlers.py#L71

That results in `ProcessorFormatter` formatting the same message twice, but the second time, it is not aware if the message itslef is a dict, as it arrives as string.

Here we are handling exactly this case - we check if the string record message that arrives, has json structure, and if it does, we treat it, as if we would treat a dict

### DIFF
--- a/src/structlog/stdlib.py
+++ b/src/structlog/stdlib.py
@@ -11,6 +11,7 @@ See also :doc:`structlog's standard library support <standard-library>`.
 
 from __future__ import absolute_import, division, print_function
 
+import json
 import logging
 
 from structlog._base import BoundLoggerBase
@@ -407,15 +408,12 @@ class ProcessorFormatter(logging.Formatter):
         """
         Extract ``structlog``'s `event_dict` from ``record.msg`` and format it.
         """
+        should_format = False
         if isinstance(record.msg, dict):
-            # Both attached by wrap_for_formatter
-            logger = record._logger
-            meth_name = record._name
-
-            # We need to copy because it's possible that the same record gets
-            # processed by multiple logging formatters.  LogRecord.getMessage
-            # would transform our dict into a str.
-            ed = record.msg.copy()
+            logger, meth_name, ed = self.get_logger_name_and_msg(record)
+        elif self.is_json(record.msg):
+            record.msg = json.loads(record.msg)
+            logger, meth_name, ed = self.get_logger_name_and_msg(record)
         else:
             logger = None
             meth_name = record.levelname.lower()
@@ -427,9 +425,11 @@ class ProcessorFormatter(logging.Formatter):
                 ed = proc(None, meth_name, ed)
 
             del ed["_record"]
+            should_format = True
 
         record.msg = self.processor(logger, meth_name, ed)
-        return super(ProcessorFormatter, self).format(record)
+        msg = super(ProcessorFormatter, self).format(record) if should_format else record.msg
+        return msg
 
     @staticmethod
     def wrap_for_formatter(logger, name, event_dict):
@@ -444,3 +444,23 @@ class ProcessorFormatter(logging.Formatter):
         configuration.
         """
         return (event_dict,), {"extra": {"_logger": logger, "_name": name}}
+
+    @staticmethod
+    def get_logger_name_and_msg(record):
+        # Both attached by wrap_for_formatter
+        logger = record._logger if hasattr(record, '_logger') else None
+        meth_name = record._name if hasattr(record, '_name') else None
+
+        # We need to copy because it's possible that the same record gets
+        # processed by multiple logging formatters.  LogRecord.getMessage
+        # would transform our dict into a str.
+        ed = record.msg.copy()
+        return logger, meth_name, ed
+
+    @staticmethod
+    def is_json(message):
+        try:
+            json.loads(message)
+        except ValueError:
+            return False
+        return True


if-fi:
@hynek The proposed code change, works very well on on my side, but I see one test is failing. 
I will try to figure out why, tomorrow, in the meantime, please have a look at the diff

src/structlog/stdlib.py
```diff
@@ -407,15 +408,12 @@ def format(self, record):
...
-            # Both attached by wrap_for_formatter
-            logger = record._logger
-            meth_name = record._name
-
-            # We need to copy because it's possible that the same record gets
-            # processed by multiple logging formatters.  LogRecord.getMessage
-            # would transform our dict into a str.
-            ed = record.msg.copy()
+            logger, meth_name, ed = self.get_logger_name_and_msg(record)
+        elif self.is_json(record.msg):
```
gilbsgilbs: Perfomance-wise that must be really slow and conceptually, that does not look ideal. What if I decide to log a json-valid string ? I don't think there are any reason to treat this type of message differently. Moreover, if it's actually a json-string, you're parsing it twice. The proper way to do it would be a `try/except` in the else clause.

Not to appear closed-minded, but I reckon that the proper fix to this issue is once again something closer to what I suggested and started to implement in https://github.com/hynek/structlog/pull/115#issuecomment-300491382 . ProcesserHandler would wrap your Handler seamlessly. I do think that relying on the formatter to do this job wasn't the best idea in the first place. :-° But there's probably another, more elegant, more performant workaround.

src/structlog/stdlib.py
```diff
@@ -444,3 +444,23 @@ def wrap_for_formatter(logger, name, event_dict):
         configuration.
         """
         return (event_dict,), {"extra": {"_logger": logger, "_name": name}}
+
+    @staticmethod
+    def get_logger_name_and_msg(record):
+        # Both attached by wrap_for_formatter
+        logger = record._logger if hasattr(record, '_logger') else None
```
hynek: JFTR, `hasattr()` is an [unfortunate Python idiom](https://hynek.me/articles/hasattr/) and `getattr(record, "_logger", None)` does *exactly* what you want here; just faster and better. :)

src/structlog/stdlib.py
hynek: Yeah sorry, this is unfortunately no bueno. :(  We had similar things proposed before but having a parser part of logging is not acceptable.

hynek:
(continuing from the comment so it’s on record)

Unfortunately I don’t have time/mental cycles to dive into #115 ATM (flying off to PyCon US in less than two days).  So I guess I’ll try to push out 17.2 with #117 and #118 and then look at the better but more complex fixes.  I think I’ll also add a note about #119 for now.  But more’s not possible in short term. :(

I may dig into it at PyCon US sprints tho.

src/structlog/stdlib.py
if-fi: @gilbsgilbs please check out #119's description and the whole commit message. There it is explained in detail what the problem is.

if-fi:
@hynek #119 is does not occur in version 17.2.0
You can close this pull request

hynek:
Huh, really!?

if-fi:
Yep. I upgraded to 17.2.0 about a week ago and since then I haven't seen the problem described in the ticket on any environment

Issue #138, opened 2017-08-12
Using renderers as standard library formatters


quodlibetor:
Given the large amount of existing code that depends on the standard library's logging, it would be really nice to be able to use structlog both for constructing log events (which it excels at) *and* for formatting existing events.

To support this, it seems like it would be awesome if `*Renderer`s had some method that transforms them into the inverse of the `render_to_log_kwargs` processor.

That would allow me to do something like:

```python
structlog.configure(
    processors=[
        ...
        structlog.stdlib.render_to_log_kwargs,
    ],
    ...
)

my_handler = get_handler()
if DEV:
    my_handler.setFormatter(structlog.dev.ConsoleRenderer().as_formatter())
else:
    my_handler.setFormatter(structlog.JsonRenderer().as_formatter())
logging.getLogger().addHandler(my_handler)
```

I'm not sure if this is something that is currently supported and I just can't figure out how to do it, or how much effort it would take to make it happen.

hynek:
Did you have a look at http://www.structlog.org/en/stable/standard-library.html?  There's plenty ways to integrate structlog with stdlib.

quodlibetor:
Gah, yes, but I didn't look *correctly*. From what I can see the `ProcessorFormatter` is exactly what I want, and is well documented.

PR #140, opened 2017-08-17
Add docs around using ProcessorFormatter

quodlibetor:
This adds a (I think) gentler introduction to the way that ProcessorFormatter
works. The existing dictConfig remains. I wasn't sure if I should maybe add
another couple of headings to the docs to make it more clear, we're already at
the maximum nesting level in sphinx so it won't show up in the ToC anyway.

I think that if this had existed I wouldn't have opened #138


hynek:
Thanks this looks great, I’ll some little polishing but I’m merging it for now.

FWIW, I intend to redo structlog's docs a bit in the vein of  <https://www.divio.com/en/blog/documentation/>.

PR #141, opened 2017-08-29
Send colored logs to non-tty destinations + fix for a unit test

sb98052:
Add option to ConsoleRenderer for sending colored logs to non-tty destinations. 
The main use case is that of files meant to be streamed to the console. 
We use this with docker logs.
Also included is a fix for a unit test.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,8 @@ Changes:
 - Empty strings are valid events now.
   `#110 <https://github.com/hynek/structlog/issues/110>`_
 
+- ``structlog.dev.ConsoleRenderer`` now accepts a *force_colors* argument to output colored logs even if the destination is not a tty. 
+  Use this option if your logs are stored in files that are intended to be streamed to the console.
 
 ----
 
--- a/tests/test_dev.py
+++ b/tests/test_dev.py
@@ -228,6 +228,29 @@ class TestConsoleRenderer(object):
         assert dev._PlainStyles is plain_cr._styles
         assert "[info     ] event                          foo=bar" == rv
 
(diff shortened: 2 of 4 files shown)


hynek:
This looks useful however CI is failing (I think you just have to split a line) and it needs a changelog entry.

thanks!

(p.s. I’ll be off the grid in the next two weeks so it might take a while until I'll be able to re-review.)

CHANGELOG.rst
```diff
@@ -27,6 +27,7 @@ Changes:
 - Empty strings are valid events now.
   `#110 <https://github.com/hynek/structlog/issues/110>`_
 
+- `structlog.dev.ConsoleRenderer`` now accepts a *force_colors* argument to output colored logs even if the destination is not a tty. Use this option if your logs are stored in files that are intended to be streamed to the console.
```
hynek: - You're missing a backtick here.
- We use [semantic newlines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/) in our reST docs.

tests/test_dev.py
```diff
@@ -228,6 +228,30 @@ def test_colorama_colors_false(self):
         assert dev._PlainStyles is plain_cr._styles
         assert "[info     ] event                          foo=bar" == rv
 
+    @pytest.mark.skipif(not dev._has_colorama,
+                        reason="Colorama must be missing.")
```
hynek: This doesn't seem right?

sb98052:
Thanks for the review. It appears that CI is broken even without this change set. I debugged the issue and have submitted a proposed fix as part of this pull request.

hynek:
Hi, I’m super sorry it took so long!  Thank you for your contribution!

PR #143, opened 2017-08-31
[DOCS] fix threadlocal docs to provide correct example

maiamcc:
Threadlocal docs currently say to use an _instance_ of `WrappedDictClass` as the `context_class`, i.e.:
```
configure(context_class=WrappedDictClass())
```

The call above doesn't actually set `context_class` at all, b/c an empty `WrappedDictClass` instance evaluates to falsey and so fails the `if context_class` check in [`configure()`](https://github.com/hynek/structlog/blob/master/src/structlog/_config.py#L164). If we manage to actually set `context_class` as an _instance_ rather than a _class_, attempts to `.bind()` result in error: 
```
TypeError: 'WrappedDict-f8fcd814-393b-4ddf-b539-c55e7bd04e62' object is not callable
```

Worth noting that the [threadlocal tests](https://github.com/maiamcc/structlog/blob/master/tests/test_threadlocal.py#L37) all test behavior with a `wrap_dict` _class_, not _instance_.

### DIFF
--- a/docs/thread-local.rst
+++ b/docs/thread-local.rst
@@ -61,13 +61,12 @@ Within one thread, every instance of the returned class will have a *common* ins
    <WrappedDict-...({'a': 1, 'b': 2, 'c': 3})>
 
 
-Then use an instance of the generated class as the context class::
+To enable thread local context, then, use the generated class as the context class::
 
-   configure(context_class=WrappedDictClass())
+   configure(context_class=WrappedDictClass)
 
 .. note::
-   **Remember**: the instance of the class *doesn't* matter.
-   Only the class *type* matters because *all* instances of one class *share* the *same* data.
+   Creation of a new ``BoundLogger`` initializes the logger's context as ``context_class(initial_values)``, and then adds any values passed via ``.bind()``. As all instances of a wrapped dict-like class share the same data, in the case above, the new logger's context will contain all previously bound values in addition to the new ones.
 
 :func:`structlog.threadlocal.wrap_dict` returns always a completely *new* wrapped class:
 


maiamcc:
Also, would you be open to a PR to change the checks in [`configure()`](https://github.com/hynek/structlog/blob/master/src/structlog/_config.py#L164) from `if not <arg>` to `if <arg> is not None`? Presumably what we want to know is whether or not the user has passed an arg (i.e. whether it has changed from its default value of `None`) , NOT whether the arg is falsey. Like, it probably shouldn't be falsey, but imo better behavior is to accept the falsey arg and throw an error later in the process, when we try to do something with that arg.

The current behavior of `configure()` gave me a huge headache when I followed the docs and tried to configure `context_class=WrappedDictClass()` -- it wasn't throwing any errors _because `configure()` rejected my arg` since it was falsey. It wasn't until I tried that same `configure()` call after my wrapped dict contained things and was no longer falsey that I ran into the `TypeError: ... not callable` mentioned above. Would have been much easier to debug if I got that `TypeError` as soon as I tried to use a wrapped dict instance as context class.

hynek:
JFTR, this is terrible timing I've just landed in South Africa and will be pretty much off the grid for almost three weeks.  Please be patient. :)

maiamcc:
No worries, take your time! Enjoy your trip :)

hynek:
Thanks!  OK I’m back…trying to remember what that Python thingie was… ;)

I’ll merge and fix it, but JFTR for the next time, we use [semantic new lines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/) in the docs. :)

***

And yes, bare if-statements are terrible and I should feel bad about them. 🙈  Feel free to shoot me a PR, otherwise I’ll fix it myself eventually™.

PR #145, opened 2017-10-14
twisted: don't encapsulate failures twice

### LINKED ISSUE #144: Failure in TestExtractStuffAndWhy with Twisted 17.9
Hello,

Following bug [#876988 in Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=876988), when I try to run tests with Twisted 17.9 installed, I get the following error:

```
================================================= FAILURES =================================================
_______________________________ TestExtractStuffAndWhy.test_handlesFailures ________________________________

self = <tests.test_twisted.TestExtractStuffAndWhy object at 0x7f2a8a4917d0>

    def test_handlesFailures(self):
        """
            Extracts failures and events.
            """
        f = Failure(ValueError())
>       assert (
            (f, "foo", {}) ==
            _extractStuffAndWhy({"_why": "foo",
                                 "_stuff": f})
        )
E       AssertionError: assert (<twisted.pyt... >, 'foo', {}) == (<twisted.pyth... >, 'foo', {})
E         At index 0 diff: <twisted.python.failure.Failure exceptions.ValueError: > != <twisted.python.failure.Failure exceptions.ValueError: >
E         Full diff:
E         (<twisted.python.failure.Failure exceptions.ValueError: >, 'foo', {})

tests/test_twisted.py:117: AssertionError
============================= 1 failed, 229 passed, 10 skipped in 0.50 seconds =============================
```

Putting some print arounds, the problem is that `isinstance(_stuff, BaseException)` is true, so we create a new encapsulate `_stuff` twice. I'll propose a fix shortly.

vincentbernat:
If a failure is already a Failure instance, don't encapsulate it into
a Failure. Keep it as is.

Fix #144 

Note that I have no clue if this is actually correct. This fixes the failing test and it doesn't seem dumb to do that. I didn't investigate what changes triggered this problem. Maybe `Failure` was not a `BaseException` in the past.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,8 @@ matrix:
       env: TOXENV=py27-colorama
     - python: "3.6"
       env: TOXENV=py36-colorama
+    - python: "3.6"
+      env: TOXENV=py36-oldtwisted
 
 
     # Meta
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,7 +26,8 @@ Changes:
 
 - Empty strings are valid events now.
   `#110 <https://github.com/hynek/structlog/issues/110>`_
-
+- Do not encapsulate Twisted failures twice with newer versions of Twisted.
+  `#144 <https://github.com/hynek/structlog/issues/144>`_
 
 ----
 
--- a/src/structlog/twisted.py
+++ b/src/structlog/twisted.py
@@ -107,7 +107,7 @@ def _extractStuffAndWhy(eventDict):
         _stuff = Failure()
     # Either we used the error ourselves or the user supplied one for
     # formatting.  Avoid log.err() to dump another traceback into the log.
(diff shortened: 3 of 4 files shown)


hynek:
So yeah looks fair enough, however I think we have to add a tox env for Twisted with the old behavior to ensure we don’t break anything?  It should be enough if it runs just the twisted  parts of the test suite.

vincentbernat:
Do you want me to do it?

hynek:
Yeah it would be nice if it were part of this PR.  Also please remember to adapt .travis.yml.

vincentbernat:
Done. Error seems unrelated to this PR.

hynek:
Thanks, could you add a changelog entry too, please?

vincentbernat:
I have modified the first commit to add a changelog entry.

hynek:
Thanks!

PR #146, opened 2017-10-17
Skip simplejson test if module is not installed.

piotr1212:


hynek:
Please hang on, we’ll have to figure out #144 first. :|

hynek:
thanks!

PR #147, opened 2017-12-11
lazy evaluation of args for _proxy_to_logger

gsemet:
Allow the smooth transition of module from logging to structlog

Use case:
- the developer has a module that uses `logging`, and also likes lazy evaluation of arguments, for example:

    log = logging.getLogger(__name__)
    logging.debug("my value is %s", myvalue)

- He wishes to transition to `structlog`, so he replaces by 

    from structlog import get_logger
    log = get_logger()

- the developer does not want to modify anything else
 
In unit test it may fail with the following error:
```
TypeError: _proxy_to_logger() takes from 2 to 3 positional arguments but 4 were given
```

Because `BoundLoggerBase._proxy_to_logger` does not accept `*args` by default.

With this patch, we allow transparently to have lazy evaluated arguments, allowing the use of 

    log = structlog.get_logger()
    logging.debug("my value is %s", myvalue)

Signed-off-by: Gaetan Semet <gaetan@xeberon.net>

### DIFF
--- a/src/structlog/_base.py
+++ b/src/structlog/_base.py
@@ -149,7 +149,7 @@ class BoundLoggerBase(object):
                 "string."
             )
 
-    def _proxy_to_logger(self, method_name, event=None, **event_kw):
+    def _proxy_to_logger(self, method_name, event=None, *args, **event_kw):
         """
         Run processor chain on event & call *method_name* on wrapped logger.
 
@@ -162,6 +162,8 @@ class BoundLoggerBase(object):
             user called because it also get passed into processors.
         :param event: The event -- usually the first positional argument to a
             logger.
+        :param args: argument to render into the event (for compatibility with
+            lazy argument evaluation of logging.* functions)
         :param event_kw: Additional event keywords.  For example if someone
             calls ``log.msg('foo', bar=42)``, *event* would to be ``'foo'``
             and *event_kw* ``{'bar': 42}``.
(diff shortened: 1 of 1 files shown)


hynek:
Hmmmm, I feel like this is an already solved problem…have you seen [structlog.stdlib.PositionalArgumentsFormatter](http://www.structlog.org/en/stable/api.html#structlog.stdlib.PositionalArgumentsFormatter)?

hynek:
Well it does, if you run `structlog.configure()` somewhere.

I can’t imagine a project that can run solely on default values (and as the docs point out: they tend to change – so it’s dangerous too), so it seems like these changes would buy a user a day at best, until they realise they have to configure anyways?

Stibbons:
I would give the example of unit test. My project of course using structlog.configure, and it works great. However, I have some unit test (using pytest) that execute and came accross a module that use `structlog.get_logger`, but now logs are captured by pytest so they probably use their own logging configuration (I have not investigated).

hynek:
That sounds like a bug in your test setup.  I have just added `structlog.get_config()`  to master that should help you to find the culprit!

FWIW, I usually configure structlog within `pytest_configure` in conftest.py (that should *not* be in your `tests` directory) and haven’t ever run in such problems. Another option is an `autouse=True` fixture.

Stibbons:
This mainly occurs on project ("ProjectName") that depends on a project that uses structlog ("DependencyName"). DependencyName uses structlog, ProjectName only use logging.
So structlog's logger needs to behave by default like `logging.getLogger`

PR #151, opened 2018-02-03
Add log level numbers

goodspark:
When combing through logs, I find it's sometimes helpful to filter by a range of log levels. Currently this must be done by checking log level strings. This adds the number so it's easier to do such a check:

With only log levels:
`log_level in ("error", "warning", "critical")`

With log level numbers:
`log_level_number >= 30`

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -187,6 +187,8 @@ API Reference
 
 .. autofunction:: filter_by_level
 
+.. autofunction:: add_log_level_number
+
 .. autofunction:: add_log_level
 
 .. autofunction:: add_logger_name
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -187,9 +187,9 @@ API Reference
 
 .. autofunction:: filter_by_level
 
-.. autofunction:: add_log_level_number
-
 .. autofunction:: add_log_level
+
+.. autofunction:: add_log_level_number
 
 .. autofunction:: add_logger_name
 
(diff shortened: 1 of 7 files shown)


goodspark:
Looks like style failures are from changes I haven't made. Would you prefer they be fixed in this PR or separately and then this branch merged/rebased?

docs/api.rst
```diff
@@ -187,6 +187,8 @@ API Reference
 
 .. autofunction:: filter_by_level
 
+.. autofunction:: add_log_level_number
```
hynek: I think this should go under add_log_level since that is still the main one.

docs/standard-library.rst
```diff
@@ -58,6 +58,9 @@ Processors
 :func:`~structlog.stdlib.add_logger_name`:
    Adds the name of the logger to the event dictionary under the key ``logger``.
 
+:func:`~structlog.stdlib.add_log_level_number`:
```
hynek: Same as above, maybe explain a bit more what a log level number is? Someone reading this who doesn’t know what a log level number is, will have not clue what this is support to be.

src/structlog/stdlib.py
```diff
@@ -330,6 +330,14 @@ def filter_by_level(logger, name, event_dict):
         raise DropEvent
 
 
+def add_log_level_number(logger, method_name, event_dict):
+    """
+    Adds the log level number to the event dict.
```
hynek: Kind of as above: it should broadly explain what that is, maybe even link to the docs.

Also: `Add` not `Adds` please.

goodspark:
@hynek Take another look? I believe I've addressed issues

hynek:
Thanks!

PR #174, opened 2018-07-06
Added brackets to list comprehension, fixing this exception:

ricardopinto:
```
Exception ignored in: <generator object _find_first_app_frame_and_name.<locals>.<genexpr> at 0x10693ad00>
  File "/Users/ricardopinto/a_project/env/lib/python3.6/site-packages/structlog/_frames.py", line 42, in <genexpr>
    while any(name.startswith(i) for i in ignores):
SystemError: error return without exception set
```

From what I saw, `ignores` was `['structlog', 'logging']`, and `name` was `structlog._frames`.
Python version is 3.6.5, structlog is 18.1.0.

### DIFF
--- a/src/structlog/_frames.py
+++ b/src/structlog/_frames.py
@@ -38,7 +38,7 @@ def _find_first_app_frame_and_name(additional_ignores=None):
     ignores = ["structlog"] + (additional_ignores or [])
     f = sys._getframe()
     name = f.f_globals.get("__name__") or "?"
-    while any([name.startswith(i) for i in ignores]):
+    while any(tuple(name.startswith(i) for i in ignores)):
         if f.f_back is None:
             name = "?"
             break
--- a/src/structlog/_frames.py
+++ b/src/structlog/_frames.py
@@ -38,7 +38,7 @@ def _find_first_app_frame_and_name(additional_ignores=None):
     ignores = ["structlog"] + (additional_ignores or [])
     f = sys._getframe()
     name = f.f_globals.get("__name__") or "?"
-    while any(name.startswith(i) for i in ignores):
+    while any([name.startswith(i) for i in ignores]):
         if f.f_back is None:
             name = "?"
             break


hynek:
I’d need to see a breaking test for the old code.

IMHO the code itself should be fine: it’s generator which is more effective than a list comprehension but if it’s broken, we need a regression test to make sure the error doesn’t come back.

Out of curiosity, is numpy involved? Googling that error brings up a bunch of it.

hynek:
Looks like we’re not the only ones with this problem: https://github.com/requests/requests/issues/4567

It seems like there’s something weird going on with frames, C extensions and threads. The unholy trifecta. :)

Would you mind changing `[]` to `tuple()` and see if it helps too? They’re a bit more efficient.

techdragon:
I'm getting quite a lot of log noise due to exceptions in this generator, so fixing this would be very welcome.

hynek:
@techdragon would you mind to test whether the proposed solution with `tuple()` instead of [] works then?

ricardopinto:
@hynek I tried `tuple()` instead of `[]` and it worked. Want me to update the code in the PR?

hynek:
That would be swell!

ricardopinto:
@hynek done. Happy new year!

hynek:
thanks and sorry for the delays!

ricardopinto:
Excellent, thanks!

methane:
@ricardopinto Sorry for asking about old issue, but do you remember that you did use PyCharm?

It seems there was a bug in PyCharm and python raised SystemErro during genexpr.

https://stackoverflow.com/questions/50369959/systemerror-error-return-without-exception-set-when-using-requests-and-debugge/52054289#52054289

https://stackoverflow.com/questions/50055316/flask-upgrade-lots-of-exception-ignored-in-messages-in-console

ricardopinto:
> @ricardopinto Sorry for asking about old issue, but do you remember that you did use PyCharm?
> 
> It seems there was a bug in PyCharm and python raised SystemErro during genexpr.
> 
> https://stackoverflow.com/questions/50369959/systemerror-error-return-without-exception-set-when-using-requests-and-debugge/52054289#52054289
> 
> https://stackoverflow.com/questions/50055316/flask-upgrade-lots-of-exception-ignored-in-messages-in-console
> 
> https://github.com/ContinuumIO/anaconda-issues/issues/8737

Hi. Yes, I was using PyCharm when I got that error.

methane:
Thank you. I am relieved to hear that.
There is no unsolved multithread+genexpr issue in cpython.

PR #188, opened 2018-11-23
Check a dict passed to ProcessorFormatter actually came from structlog.

### LINKED ISSUE #187: ProcessorFormatter doesn't handle librarys which send dicts to standard logger well
The processor formatter uses this line to distinguish between Structlog and regular messages:
```
if isinstance(record.msg, dict):
```
Unfortunately this runs afoul of any library or bug which happens to accidentally pass a `dict` object to the Python standard library logger (in the wild I've found the [connexion](https://github.com/zalando/connexion) library does this in some of its initialization code, producing exceptions when using structlog). This happens because the `record` entry doesn't have `_logger` attached because it didn't come through structlog.

IMO the ideal solution would be for the `wrap_for_formatter` handler to actually supply a wrapped dictionary type which could be tested for directly, but this doesn't seem particularly viable.

The best solution therefore is probably to embrace duck-typing in this code path and change the check to ensure the expected attributes are actually there:
```
if isinstance(record.msg, dict) and hasattr(record, "_logger") and hasattr(record,"_name"):
```

wrouesnel:
ProcessorFormatter assumes certain attributes are attached by wrap_for_formatter
but originally only tested if it was receiving a dictionary as the log message.

This breaks if any libraries you use happen to pass dictionaries as log messages
expecting them to print using default string formatting.

Fortunately since the behavior of the LogRecord is different between the two
paths, we can resolve this by testing for the relevent attrs before choosing.

Fixes #187


src/structlog/stdlib.py
```diff
@@ -451,7 +451,11 @@ def format(self, record):
         # Make a shallow copy of the record to let other handlers/formatters
         # process the original one
         record = logging.makeLogRecord(record.__dict__)
-        if isinstance(record.msg, dict):
+        if (
+            isinstance(record.msg, dict)
+            and hasattr(record, "_logger")
```
hynek: Please meet: https://hynek.me/articles/hasattr/ :)

If we pile on checks and assume that dicts are predominant, why don't we invert the whole thing and make it an `try:…except (KeyError, AttributeError)`? I reckon that’s considered more pythonic too.

wrouesnel: I would argue when you have the ProcessorLogger in the mix, dicts and strs aren't predominant one way or the other, so a try...except seems unnecessarily slow.

I wonder if instead the wrapper should convert LogRecord into WrappedLogRecord and just do a type-check since ultimately what we're looking for is a different object type.

hynek: >  try...except seems unnecessarily slow.

Hm, do you really think it's slower than multiple checks on each record? I think I'd like to see a benchmark on that since this is is hot code. If you have some spare it would be great if you could compare it using https://pypi.org/project/perf/ . I only care about modern Python versions FWIW.

> I wonder if instead the wrapper should convert LogRecord into WrappedLogRecord and just do a type-check since ultimately what we're looking for is a different object type.

That sound clean, but really slow? 🤔

Sorry I'm being so anal about performance here but I take great pride that structlog isn't a handbrake like logging and I'd like it to stay that way. 😇

wrouesnel: No I agree about keeping it quick. I'll setup some tests.

hynek: Awesome thank you and sorry for the delay. My original plan was to run them myself but just didn't get around to it.

hynek:
bump? :)

hynek:
So I decided to use the approach from #189 but I added your test case using you as an author: 44ccae2a63f22f52755429780b1feaeb5b9c455e

Thanks!

PR #189, opened 2018-11-30
Make ProcessorFormatter able to log dict

ZlatyChlapec:
Fixes error when logging dict via std library logger using `ProcessorFormatter`.


tests/test_stdlib.py
```diff
@@ -495,9 +495,9 @@ def test_foreign_delegate(self, configure_for_pf, capsys):
 
         assert ("", "foo [in test_foreign_delegate]\n") == capsys.readouterr()
 
-    def test_clears_args(self, capsys, configure_for_pf):
+    def test_clears_args(self, configure_for_pf, capsys):
```
ZlatyChlapec: I changed this so it's according to style in whole class but I'll remove the change if that's a problem.

tests/test_stdlib.py
```diff
@@ -495,9 +495,9 @@ def test_foreign_delegate(self, configure_for_pf, capsys):
...
         """
-        We render our log records before sending it back to logging.  Therefore
+        We render our log records before sending it back to logging. Therefore
```
ZlatyChlapec: I guess that double space was mistake. If that's not the case I'll remove the change as well.

ZlatyChlapec:
I am not sure what to do about failed CI. Looks like it just failed to execute.

hynek:
Hehe this is an almost 1:1 dupe of #188. :) Double spacing is on purpose btw. There's mixed opinions on it, I personally usually double space in fixed-width context like source code..

ZlatyChlapec:
My bad I didn't check if there isn't anything about it before making PR. I changed it to `try..except`. Feel free to close in favor of #188.

hynek:
Looks like #188 is stalled for now. Would you like to finish this up?

ZlatyChlapec:
Definitely, I would love that.

hynek:
Ah you know what? Nothing. :) I prefer the try/except way anyways and I'll fix the minor stuff myself. Thank you for your contribution and sorry it took so long!

PR #195, opened 2019-02-21
Make filter_by_level processor work with std lib logs

ZlatyChlapec:
When using `filter_by_level` processor with `ProcessorFormatter` it fails on std lib logs because of https://github.com/hynek/structlog/blob/c0aeab5f3ada09b7079d7ba7dc66314e1d8d2792/src/structlog/stdlib.py#L333 being called on `None`.

### DIFF
--- a/src/structlog/stdlib.py
+++ b/src/structlog/stdlib.py
@@ -461,7 +461,7 @@ class ProcessorFormatter(logging.Formatter):
             # would transform our dict into a str.
             ed = record.msg.copy()
         except AttributeError:
-            logger = None
+            logger = logging.getLogger(record.name)
             meth_name = record.levelname.lower()
             ed = {"event": record.getMessage(), "_record": record}
             record.args = ()
@@ -483,7 +483,7 @@ class ProcessorFormatter(logging.Formatter):
             # Non-structlog allows to run through a chain to prepare it for the
             # final processor (e.g. adding timestamps and log levels).
             for proc in self.foreign_pre_chain or ():
-                ed = proc(None, meth_name, ed)
+                ed = proc(logger, meth_name, ed)
 
             del ed["_record"]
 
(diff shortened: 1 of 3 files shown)


hynek:
Can I haz changelog?

src/structlog/stdlib.py
```diff
@@ -461,7 +461,7 @@ def format(self, record):
             # would transform our dict into a str.
             ed = record.msg.copy()
         except AttributeError:
-            logger = None
+            logger = logging.getLogger(record.name)
```
hynek: Is there a better way to achieve this? It looks expensive to call getLogger on each logging call.

ZlatyChlapec: I've got three ideas in my mind:
1. We could make it so that users would have to manually set logger to `ProcessorFormatter` and then make `ProcessorFormatter` to take value from there.
```python
logger = logging.getLogger()
formatter = ProcessorFormatter(JSONRenderer())

handler = StreamHandler(sys.stdout)
handler.setFormatter(formatter)
handler.formatter.logger = logger
# Or via setter which we would create handler.formatter.set_logger(logger)

logger.addHandler(handler)
```

2. Make users patch `logging.Logger.makeRecord` so that it includes `self` as `_logger`.

3. Create some kind of first use only cache (probably some dict with `key` name of logger and `value` logger itself) which would look up logger only first time, store it into cache and take it from there next time.

Do you have anything better in mind?

rasto2211: I'd prefer 1). ProcessorFormatter could take the `logger` as a constructor argument.

rasto2211:
Is anyone working on this issue?

hynek:
I'm afraid I dropped the ball here unfortunately. You still game @ZlatyChlapec ?

hynek:
fixed in #219, thanks anyway and sorry for my failure to respond in adequate time!

PR #207, opened 2019-05-03
event key parameter for ConsoleRender

mplachter:
Add Named Argument for ConsoleRender incase user creates Processor to rename the `event` key to another name they can define what the new `key` to `Render` correctly into the `Console`

### DIFF
--- a/src/structlog/dev.py
+++ b/src/structlog/dev.py
@@ -111,6 +111,8 @@ class ConsoleRenderer(object):
         must be a dict from level names (strings) to colorama styles. The
         default can be obtained by calling
         :meth:`ConsoleRenderer.get_default_level_styles`
+    :param string event_key_name: The name of the event key to be used out of
+        the event dict. the default value is set to 'event'.
 
     Requires the colorama_ package if *colors* is ``True``.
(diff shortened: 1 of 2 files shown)


mplachter:
Looks like `Travis` failed because of `Black`

https://github.com/ambv/black.

mplachter:
`Black` doesn't fail locally.

```
black....................................................................Passed
flake8...................................................................Passed
seed isort known_third_party.............................................Passed
isort....................................................................Passed
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Debug Statements (Python)................................................Passed
```

mplachter:
Found the issue was with the `line-length` `black` was working at expected. No worries about the above.

hynek:
Could you elaborate on the reasoning behind this?  I could be wrong, but there is no precedent in structlog's APIs for that, is there?

mplachter:
@hynek if i create a `processor` to change the `event` key to another key let's say `message` and then try to use the built in `ConsoleRender` which explicitly looks for the `event` key it will throw an error.

PR #214, opened 2019-05-23
add proof-of-concept classes for easy logging to Graylog using GELF

danchr:
The documentation on [Logging Best Practices](https://www.structlog.org/en/stable/logging-best-practices.html) states this:

> _Graylog_ goes one step further.
> It not only supports everything those above do (and then some); you can also directly log JSON entries towards it — optionally even through an AMQP server (like _RabbitMQ_) for better reliability. 
> Additionally, _Graylog's Extended Log Format_ (GELF) allows for structured data which makes it an obvious choice to use together with ``structlog``.

After some searching, I was unable to find something that made this as easy at it should be. Some packages had done some of the work, but they either seemed unaware of `structlog`, or didn't integrate with the regular `logging` infrastructure as used in e.g. Django.

This is an attempt to remedy this. This PR contains neither documentation nor tests, but exists mostly to gauge interest. Is this something you'd want in `structlog`?

### DIFF
--- a/src/structlog/gelf.py
+++ b/src/structlog/gelf.py
@@ -1,0 +1,244 @@
+"""Helper classes for logging to Graylog using GELF"""
+
+import datetime
+import json
+import logging.handlers
+import math
+import socket
+import struct
+import zlib
+
+try:
+    import pika
+except ImportError:  # nocov
+    pika = None
+
+
+class GELFFormatter(logging.Formatter):
+    """Format records as binary GELF.
+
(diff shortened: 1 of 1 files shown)


hynek:
This is _way_ too much code that I never use, so I really can't add it to structlog proper.

If you're using it, it would be really cool if you could publish the code as `structlog-gelf` and link it from our [wiki](https://github.com/hynek/structlog/wiki/Third-Party-Extensions)!

Issue #173, opened 2018-07-06
Traceback missing


md1023:
My Django 1.8, Python 2 application uses third party middleware:
```
import struct
logger = structlog.get_logger(__name__)

class MyMiddleware(AuthMiddleware):
    def process_view(self, *args, **kwargs):
        print('FOO')
        1/0
        return super(MyMiddleware, self).process_view(*args, **kwargs)
```

My output misses tracebacks:
```
FOO
2018-07-06 11:06.36 {'events': []}
```
I can't debug, even see errors!

I did try zero-config options, other configuration options, `structlog.processors.ExceptionPrettyPrinter()`, but couldn't find who swallows and exception. I looked for try/excepts with pass, but there's not one for `1/0`.

Please suggest some idea, I'm running out of debug options.

md1023:
I forgot Django does some 'magic' and captures logs.

I put extra logger:
```
    'loggers': {
        '': {
            'handlers': ['colored_console'],
            'level': 'WARNING',
        },
        'django': {
            'handlers': ['colored_console'],
            'level': 'WARNING',
        },
```
To my excitement, exceptions did appear. Unfortunately not expected ones:
```
  File "/usr/local/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/grant_types/resource_owner_password_credentials.py", line 101, in create_token_response
    if not self.request_validator.authenticate_client(request):
  File "./uauth_server/oauth2/validators.py", line 151, in authenticate_client
    logger.exception(exc)
  File "/usr/local/lib/python2.7/site-packages/structlog/_base.py", line 176, in _proxy_to_logger
    args, kw = self._process_event(method_name, event, event_kw)
  File "/usr/local/lib/python2.7/site-packages/structlog/_base.py", line 136, in _process_event
    event_dict = proc(self._logger, method_name, event_dict)
  File "/usr/local/lib/python2.7/site-packages/structlog/dev.py", line 191, in __call__
    event += self._styles.reset
TypeError: unsupported operand type(s) for +=: 'exceptions.TypeError' and 'str'
```
This time I intentionally didn't run `structlog.configure`, relying on default configuration.

Along with this issue, could I drop 'django' from loggers and still get exceptions?

hynek:
A TypeError for 1/0 looks wrong too, so I suspect there’s something else going on.

[...]

PR #216, opened 2019-05-26
Show the default processor list configuration in getting-started

AndydeCleyre:
This facilitates making small and informed changes to the default processor list before delving into deeper docs or code. Mentioned at the end of #173, though it is not addressed to that bug.

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -38,7 +38,8 @@ Here, ``structlog`` takes full advantage of its hopefully useful default setting
   Thus, it's easy to add support for logging of your own objects\ [*]_.
 - If you have `colorama <https://pypi.org/project/colorama/>`_ installed, it's rendered in nice :doc:`colors <development>`.
 
-It should be noted that even in most complex logging setups the example would still look just like that thanks to :ref:`configuration`. Using the defaults, as above, is equivalent to::
+It should be noted that even in most complex logging setups the example would still look just like that thanks to :ref:`configuration`.
+Using the defaults, as above, is equivalent to::
 
    import structlog
    structlog.configure(processors=[
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -42,12 +42,18 @@ It should be noted that even in most complex logging setups the example would st
 Using the defaults, as above, is equivalent to::
 
    import structlog
-   structlog.configure(processors=[
-       structlog.processors.StackInfoRenderer(),
-       structlog.processors.format_exc_info,
-       structlog.processors.TimeStamper(),
-       structlog.dev.ConsoleRenderer()
-   ])
+   structlog.configure(
+       processors=[
+           structlog.processors.StackInfoRenderer(),
+           structlog.processors.format_exc_info,
+           structlog.processors.TimeStamper(),
+           structlog.dev.ConsoleRenderer()
+       ],
+       wrapper_class=structlog.BoundLogger,
+       context_class=dict,
+       logger_factory=structlog.PrintLoggerFactory(),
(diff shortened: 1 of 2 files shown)


hynek:
I agree that this would be a good thing, but need some way to ensure, that it doesn't go out of date. 🤔

Also maybe mention _all_ settings?

AndydeCleyre:
Sure, I can try to add all the default settings. As for ensuring accuracy, maybe we turn the snippet into a doctest one which creates two loggers and asserts some kind of configuration equivalence?

AndydeCleyre:
So instead, we could do something like below. Three issues to consider, though:

1. It's verbose `¯\_(ツ)_/¯`
2. This use of `ConsoleRenderer` fails the doctest because `colorama` is not in the `dev`, `tests`, `docs`, or standard requirements. Should I specify `colors=False`, or test for presence of `colorama`, or should it be included in `tests` or `dev` requirements, or something else?
3. Something's funky with `TimeStamper`, such that while two instances appear to have the same `type`, those `type`s equate neither via `is` nor `==`
4 (bonus consideration!). My local tests fail the `examples.rst` batch, indicating I may not be testing in a proper environment somehow

```rst
Below, we explicitly configure ``structlog`` to use its defaults.
The result is the same as not specifying any configuration:

.. doctest::

   >>> import structlog
   >>> default_cfg = structlog.get_config()  # save defaults for comparison
   >>> structlog.configure(
   ...     processors=[
   ...         structlog.processors.StackInfoRenderer(),
   ...         structlog.processors.format_exc_info,
   ...         structlog.processors.TimeStamper(),
   ...         structlog.dev.ConsoleRenderer()
   ...     ],
   ...     wrapper_class=structlog.BoundLogger,
   ...     context_class=dict,
   ...     logger_factory=structlog.PrintLoggerFactory(),
   ...     cache_logger_on_first_use=False
   ... )
   >>> # confirm that the explicit configuration is equivalent to the default configuration:
   >>> import types
   >>> explicit_cfg = structlog.get_config()
   >>> for prop in ('wrapper_class', 'context_class', 'cache_logger_on_first_use'):
   ...     assert default_cfg[prop] is explicit_cfg[prop]
   >>> assert type(default_cfg['logger_factory']) is type(explicit_cfg['logger_factory'])
   >>> assert len(default_cfg['processors']) == len(explicit_cfg['processors'])
   >>> for dproc, eproc in zip(default_cfg['processors'], explicit_cfg['processors']):
   ...     if isinstance(dproc, types.FunctionType):
   ...         assert dproc is eproc
   ...     else:
   ...         assert type(dproc) is type(eproc)
```

AndydeCleyre:
@hynek Could you comment on the previous post?

hynek:
I'm gonna say let's go a clean and easy way:

1. all of the config
2. put a comment in _config.py that reminds that getting-started.rst needs to be updated whenever the defaults change
3. profit

docs/getting-started.rst
```diff
@@ -42,12 +42,18 @@ It should be noted that even in most complex logging setups the example would st
...
-   ])
+   structlog.configure(
+       processors=[
+           structlog.processors.StackInfoRenderer(),
+           structlog.processors.format_exc_info,
+           structlog.processors.TimeStamper(),
+           structlog.dev.ConsoleRenderer()
+       ],
+       wrapper_class=structlog.BoundLogger,
+       context_class=dict,
```
hynek: could you add a comment here explaining that it's if the runtime has ordered dicts and OrderedDict otherwise?

AndydeCleyre: Of course, thanks.

hynek:
Thanks!

PR #219, opened 2019-07-24
Add logger argument to ProcessorFormatter

rasto2211:
`filter_by_level` processor needs logger object in order to find out what is the log level. Add `logger` as an argument to `ProcessorFormatter` and then pass it through the processor chain so `filter_by_level` can use it.

Related to #195

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,9 @@ Changes:
 - Added more pass-through properties to ``structlog.stdlib.BoundLogger``.
   To makes it easier to use it as a drop-in replacement for ``logging.Logger``.
   `#198 <https://github.com/hynek/structlog/issues/198>`_
+- `ProcessorFormatter` takes logger object as an optional keyword argument.
+  This fixes a bug when using ProcessorFormatter with
+  `stuctlog.stdlib.filter_by_level`.
 
 
 ----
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -27,6 +27,9 @@ Changes:
 - Added more pass-through properties to ``structlog.stdlib.BoundLogger``.
   To makes it easier to use it as a drop-in replacement for ``logging.Logger``.
   `#198 <https://github.com/hynek/structlog/issues/198>`_
+- `ProcessorFormatter` takes logger object as an optional keyword argument.
```
hynek: This is markdown. :) Please use double backticks.

CHANGELOG.rst
```diff
@@ -27,6 +27,9 @@ Changes:
...
+  This fixes a bug when using ProcessorFormatter with
+  `stuctlog.stdlib.filter_by_level`.
```
hynek: Same, and please pull it into the line above.

src/structlog/stdlib.py
```diff
@@ -466,6 +466,8 @@ class ProcessorFormatter(logging.Formatter):
         ``True`` to keep it on the :class:`logging.LogRecord`. (default: False)
     :param bool keep_stack_info: Same as *keep_exc_info* except for Python 3's
         ``stack_info``.
+    :param logger: Logger which we want to push through the structlog processor
```
hynek: * Please add the default value
* and a `.. versionadded` tag. 
* wrap `structlog` in double ticks – that's our corporate brand. ;)
* I feel like this needs more explanation what it is actually good for? Also maybe more info in the stdlib docs so people know (how) to actually use it?
* Bonus points for adding the default to keep_stack_info too. :)

Issue #165, opened 2018-06-07
Permission to generate stubs for structlog in typeshed


waseem18:
Hi Hynek,

I would like to add stubs for structlog to https://github.com/python/typeshed, a repository used to store PEP 484 type signatures for the Python stdlib and popular third party libraries. As per PEP 484, permission is required from the library owner before merging type signatures into typeshed - https://www.python.org/dev/peps/pep-0484/#the-typeshed-repo

These stubs will be used by mypy and Pycharm. I'll start working on creating stubs after I get permission from you and do let me know if you have any suggestions.

Thanks

hynek:
Actually I would prefer if you’d contribute them to structlog proper! Packages can ship their own stubs now.

If you’d agree to do that, I’d add an AbstractBoundLogger first.

waseem18:
@hynek Good to know that packages can ship their own stubs now. Will surely go in that way.

However I didn't get what you meant by 
> If you’d agree to do that, I’d add an AbstractBoundLogger first.

hynek:
Re: AbstractBoundLogger: I suspect it’s easier to build stubs if there’s an abstract interface that fits all bound loggers.

The docs for stubs-via-package are here btw: http://mypy.readthedocs.io/en/latest/installed_packages.html

waseem18:
@hynek Thanks for the link

It would surely be efficient if there's an abstract interface for bound loggers.

So what all loggers do you think would be better to abstract out?

hynek:
So I just noticed that untypically for me, I already have a base class that all loggers derive from: `structlog.BoundLoggerBase`!

So just use that, I may or may not add an abc at some point, but it’s nothing you need to wait for.

***

One more thing: unless you can come up with a good reason, I’d say we don’t use a `.pyi` file and write the annotations into the code in Python 2-style: http://mypy.readthedocs.io/en/latest/python2.html

Let me know if you need anything or anything is unclear.

hynek:
Any progress? 😇

waseem18:
Hi @hynek 

I was on a vacation and completely forgot about this. 

I've started going through the code base and I'm loving it. I'm planning to start on type-hinting from this weekend.

haizaar:
Good day guys,
Do you know if there was any progress on this issue?

sebpiq-kialo:
@waseem18 any progress on this ?

waseem18:
Sorry folks, I couldn't get to work on it. Anyone interested can pick it up.

Thanks

[...]

PR #223, opened 2019-09-04
Add type hints

lg-kialo:
In relation to #165, here's a first take on adding types to the library.

I went for type comments instead of separate stub files with the feeling that they might be easier to maintain. 

Of course a lot of types, especially for internal modules, are not very descriptive. There's also the option of ignoring completely those modules and keeping the types only on the external API.


lg-kialo:
@hynek I would appreciate receiving some feedback on this PR, even just an estimate of when you'll be able to look at it. I just rebased with master, but it's quite some work to keep up with the changes.

hynek:
What kind of feedback would you like at this point? I haven't reviewed yet because the CI is breaking so I assumed it's still work in progress.

lg-kialo:
@hynek 

> What kind of feedback would you like at this point? I haven't reviewed yet because the CI is breaking so I assumed it's still work in progress.

About the failing builds:
- `pypy3-threads` is failing due to an internal Python header missing for PyPy. See https://github.com/python/typed_ast/issues/97. I'm really not sure there's a way around this.
- The coverage did not complete with the latest build, not sure why. It seems like it was failing though, maybe due to the type comments?

mordras:
@hynek Just wanted to ask if there's anything we can help with here. We're absolutely willing to put some more effort into this PR if there's a chance it will get merged.

hynek:
Yes, there is. The problem is that the whole topic is non-trivial and I hadn't the time and energy to dive into it in the past three weeks.  Also I need to do a release ASAP because Python 3.8 just dropped and our stdlib module needs to be upgraded. 😞 

It would be _amazing_ if you could review and test the pull request until I have the cycles to have a closer look.

hynek:
So sorry for the delays. My main issue with this – and I was trying to think up the solution myself but failed and it led to much of the delays:

As it is right now, the type information is actually worse than useless. Which is absolutely **not** your fault, but the fault of `structlog`'s design and the expressive power we get from Python's typing.

Currently `get_logger()` and the family of `bind()` methods all return `BoundLoggerBase` but that is not a useful type to actually log anything out.

The interesting part is what the _user_ gets to see like `structlog.stdlib.BoundLogger`.

Since that depends on structlog configuration, I believe the only practical way here is to make get_logger return Any and tell the users to do something like this:

```python
logger: structlog.stdlib.BoundLogger = structlog.get_logger()
```

based on the configuration they use; which I find reasonable.

That would leave the problem of bind/unbind being wrongly typed and the question is: would it be worth it to add bind/unbind proxies to stdlib.BoundLogger that just call super() for the sake of types?

I guess it could make sense to add `structlog.stdlib.get_logger()` that is correctly typed and maybe double checks the config?

Does any of that make any sense to you?

---

Relatedly, I think it's not worth it in 2020 to put so much work in making it work on Python 2 too (I know I spoke differently but everything dragged out due to my procrastination and now I think we should just go 3.5 and call it a day).

---

To be clear, I'm not asking you to do anything just yet. we need to find a useful approach first.

merwok:
> Currently `get_logger()` and the family of `bind()` methods all return `BoundLoggerBase` but that is not a useful type to actually log anything out.
> 
> The interesting part is what the _user_ gets to see like `structlog.stdlib.BoundLogger`.

But does `BoundLoggerBase` define all the attributes and methods with the same signatures as the real/concrete BoundLogger class(es) that is used at runtime?  If yes, it would be a useful type hint.  If not,  an abstract base class or a type class could be defined for use in type hints.

hynek:
No it doesn't because the actual _logging_ methods depends on the system that people are wrapping. Twisted has different logging methods than the standard library for example. It's an explicit feature of structlog to support _any_ logging frontend and backend.

It _is_ an abstract base class in a sense already: https://github.com/hynek/structlog/blob/master/src/structlog/_base.py

TBH I'd prefer to move to composition anyways which would mean that the specific bound loggers like for stdlib would provide call proxies for all the binding and unbinding and it could be properly typed (BoundLoggerBase would become a private attribute on them). That would mean one more method call but OTOH it would also make everything cleaner and allow for proper typing of the bind methods.

merwok:
> No it doesn't because the actual _logging_ methods depends on the system that people are wrapping. [...] It's an explicit feature of structlog to support _any_ logging frontend and backend.

I think this is a part of structlog that I didn’t understand!
I mostly skipped over all the wrap logger and bound logger docs, my goal was to 1) stop building strings at log sites in my app  2) send a JSON dict to log service  3) do something to transform standard logging (from web framework and other libs).

(For the record it was easy to do 1), not always easy to configure everything for 2), and I’ve looked at things like python-json-logger for 3), thinking that structlog can’t help me for things that are not calling structlog.)

Now it seems that I have to revisit my understanding of how to use structlog!
It is good news, given that stdlib logging module is used mostly everywhere, so I’ll have to re-read the docs and better configure my projects to integrate structlog config with stdlib logger and reach my goals.

antonagestam:
Since structlog only seems to have support for modern Python versions, is there any reason not to use proper annotations over Python 2-style type comments?

sbdchd:
@antonagestam I think you're correct that the supported versions going forward are 3.6 and above.

2.7 and 3.5 support was deprecated via https://github.com/hynek/structlog/pull/244

hynek:
Hi everyone, I've spent the past three days working on this and would love your feedback on #282.

hynek:
Type hints are now merged. Thanks for your input everyone.

PR #225, opened 2019-09-21
A newer, better thread-local API

### LINKED ISSUE #222: Is there a way to do threadlocal context for only certain binds?
I looked at the `threadlocal` module to solve a problem I have, but from what I can tell, it has undesirable behavior in that it changes all my existing logging code. From what I can tell, all binds become thread-global. But what I want is: "bind this request user and request ID to all messages during this request", and have all the rest of my existing logging code work as it would otherwise.

radix:
Fixes #222 

caveat: I didn't duplicate the greenlet support because it turns out greenlet.getcurrent() doesn't seem to return separate objects for different OS-level Python threads. That means that if you somehow have `greenlet` in your dependency tree, the existing thread-local stuff will just straight up not work.

If someone cares I'm sure it'd be possible to implement something that uses greenlets and threadlocal in a way that ensures that you get distinct values in different greenlets AND distinct values in different python threads.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -51,7 +51,9 @@ Changes:
   So far, the configuration proxy, ``structlog.processor.TimeStamper``, ``structlog.BoundLogger``, ``structlog.PrintLogger`` and ``structlog.dev.ConsoleLogger`` have been made pickelable.
   Please report if you need any another class ported.
   `#126 <https://github.com/hynek/structlog/issues/126>`_
-
+- Added a new thread-local API that allows binding values to a thread-local context explicitly without affecting the default behavior of bind.
+  `#222 <https://github.com/hynek/structlog/issues/222>`_,
+  `#225 <https://github.com/hynek/structlog/issues/225>`_,
 
 ----
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -51,7 +51,7 @@ Changes:
   So far, the configuration proxy, ``structlog.processor.TimeStamper``, ``structlog.BoundLogger``, ``structlog.PrintLogger`` and ``structlog.dev.ConsoleLogger`` have been made pickelable.
   Please report if you need any another class ported.
   `#126 <https://github.com/hynek/structlog/issues/126>`_
-- Added a new thread-local API that allows binding values to a thread-local context explicitly without affecting the default behavior of bind.
+- Added a new thread-local API that allows binding values to a thread-local context explicitly without affecting the default behavior of ``bind()``.
   `#222 <https://github.com/hynek/structlog/issues/222>`_,
   `#225 <https://github.com/hynek/structlog/issues/225>`_,
 
--- a/src/structlog/threadlocal.py
+++ b/src/structlog/threadlocal.py
@@ -175,9 +175,7 @@ def merge_in_threadlocal(logger, method_name, event_dict):
     Use this as your first processor in :func:`structlog.configure` to ensure
     thread-local context is included in all log calls.
     """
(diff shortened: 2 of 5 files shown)


radix:
I'm not sure why coverage isn't working. My tests should be covering everything 100% and I don't understand how to read the output of codecov.io

hynek:
This should be helpful: https://codecov.io/gh/hynek/structlog/src/222-thread-local/src/structlog/threadlocal.py#L178 Maybe you call clear before or something?

---

That said, could you make them both a try/except block?  I'm [not a fan of hasattr](https://hynek.me/articles/hasattr/) in my codebases.

Other that that, it looks super promising. When people asked me about support contextvars, something exactly like this came to my mind. Since the code is rather trivial, I suppose it doesn't make any sense to try to make something generic?

It's definitely the better approach than having everything global by default and offer temporary binds.

hynek:
OK cool. Do you need anything from me to finish this up? If you want, you can skip narrative docs for now. I can write them in a separate PR and you'll review it (timely! 🤪).

But please refer to [CONTRIBUTING.rst](https://github.com/hynek/structlog/blob/master/.github/CONTRIBUTING.rst), there's a bunch of transgressions. :)

radix:
@hynek I already started on the docs when I saw your message, so I figured I'd finish up what I have and put it up here.

Please let me know which contribution guidelines I have missed.

tests/test_threadlocal.py
```diff
@@ -262,3 +269,26 @@ def test_new_class(self, D):
         The context of a new wrapped class is empty.
         """
         assert 0 == len(D())
+
+
+class TestNewThreadLocal(object):
+    def test_bind_and_merge(self):
```
hynek: Remember Jonathan? :) Please add [test docstrings](https://jml.io/pages/test-docstrings.html).

src/structlog/threadlocal.py
```diff
@@ -162,3 +163,45 @@ def __len__(self):
...
         method = getattr(self._dict, name)
         return method
+
+
+_CONTEXT = threading.local()
+
+
+def merge_threadlocal_context(logger, method_name, event_dict):
+    """
+    A structlog processor that merges in a global (thread-local) context.
```
hynek: You can drop the structlog here.

src/structlog/threadlocal.py
```diff
@@ -162,3 +163,45 @@ def __len__(self):
...
+    The typical use-case for this function is to invoke it early in
+    request-handling code.
+    """
+    _CONTEXT.context = {}
+
+
+def bind_threadlocal(**kwargs):
+    """
+    Put keys and values into the thread-local context.
+    Use this instead of :func:`~structlog.BoundLogger.bind` when you want some
```
hynek: Empty line like in merge above please.

CHANGELOG.rst
```diff
@@ -51,7 +51,9 @@ Changes:
   So far, the configuration proxy, ``structlog.processor.TimeStamper``, ``structlog.BoundLogger``, ``structlog.PrintLogger`` and ``structlog.dev.ConsoleLogger`` have been made pickelable.
   Please report if you need any another class ported.
   `#126 <https://github.com/hynek/structlog/issues/126>`_
-
+- Added a new thread-local API that allows binding values to a thread-local context explicitly without affecting the default behavior of bind.
```
hynek:
```
``bind()``
```

 please

radix:
Thanks for the review @hynek, I've addressed your requests.

hynek:
Thanks Chris, please come back soon! ;)

PR #228, opened 2019-09-26
Allow foreign_pre_chain To Read Foreign Log Args

### LINKED ISSUE #227: Strip LogRecord.args Attribute After foreign_pre_chain Processors Run in ProcessorFormatter
First off, thank you to everyone involved in maintaining this project!

I love the chained approach it provides to building up a log message context. Great work!

I want to start using a structlog-based formatter to output logs as json for several small to medium size code bases I manage.

In particular, I want to be able to write a `processor` that takes the *keys* from the `LogRecord.args` attribute and insert them into my `event_dict`.

Right now, when the `ProcessorFormatter` class receives a log record from `logging` rather than `structlib`, it sets `LogRecord.args = ()` after generating a message [here](https://github.com/hynek/structlog/blob/0e209d95652c0f9b2943bf0334d74f4964fc7f13/src/structlog/stdlib.py#L519).

This means that none of my downstream processors defined in `foreign_pre_chain` can work with the original `LogRecord.args` keys.

Would the maintainers be open to moving the `record.args = ()` setting to *after* the `foreign_pre_chain` processors run? I have a working version locally and can add a test if it's a PR you would accept.

Speaking more broadly, would you be open to me adding some kind of `migration` docs for projects that currently use the stdlib `logging` package but would like to switch to `structlog`?

jonathanmeier5:
Resolves #227 

When migrating a project from stdlib logging to `structlog`, an easy first step is to use `ProcessorFormatter` with processors defined in `foreign_pre_chain`. These will run on every stdlib log message sent into the formatter.

However, right now there is no way to access a LogRecord's `args` in the `foreign_pre_chain` processors because they are stripped before the record is passed to the processors.

This PR introduces a `pass_foreign_args` kwarg to `ProcessorFormatter`. When set to `True`, it will inject a foreign log record's `args` attribute into the `event_dict` under the `positional_args` key before stripping them from the record.

### DIFF
--- a/src/structlog/stdlib.py
+++ b/src/structlog/stdlib.py
@@ -469,12 +469,14 @@ class ProcessorFormatter(logging.Formatter):
     :param logger: Logger which we want to push through the ``structlog``
         processor chain. This parameter is necessary for some of the
         processors like `filter_by_level`. (default: None)
-
+    :param bool pass_foreign_args: If `True`, pass a foreign log record's
+        `args` to the event_dict under `positional_args` key. (default: False)
     :rtype: str
 
     .. versionadded:: 17.1.0
     .. versionadded:: 17.2.0 *keep_exc_info* and *keep_stack_info*
     .. versionadded:: 19.2.0 *logger*
+    .. versionadded:: 19.2.0 *pass_foreign_args*
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


hynek:
Thanks!  There's a bunch of minor things but now that 3.8 has been dropped on me I need to publish fast so I'll shape it up myself.

Issue #14, opened 2013-11-10
Add call recorder for testing purposes


hynek:
Basically a processor that records all calls and allows assertions against it.  Needs #10.

aldanor:
This is a super old issue, but still -- wondering if there's a 'recommended' way of testing structlog loggers these days?.. (which isn't entirely clear from docs or issues, hence the question)

aldanor:
Here's what I'm usually doing in pytest:

```python
class LogCapture:
    def __init__(self):
        self.entries = []

    def __call__(self, _, method_name, event_dict):
        event_dict['log_level'] = method_name
        self.entries.append(event_dict)
        raise structlog.DropEvent


@pytest.fixture(scope='function')
def log_output():
    return LogCapture()


@pytest.fixture(scope='function', autouse=True)
def configure_structlog(log_output):
    structlog.configure(
        processors=[log_output]
    )
```

and then you can just

```python
def test_my_stuff(log_output):
    do_something()
    assert log_output.entries == [...]
```

hynek:
Honestly that’s pretty much I had in mind but never got around to implement it in a way that it would warrant the inclusion in structlog proper. Lately I've also been kinda lazy and just used caplog/capsys + simples checks to ensure an entry is in there. 🤔

rgalanakis:
Here's a non-pytest version, in case anyone comes across this and needs it:

```python
class LogCapture:
    def __init__(self):
        self.entries = []

    def __call__(self, _, method_name, event_dict):
        event_dict['log_level'] = method_name
        self.entries.append(event_dict)
        raise structlog.DropEvent


@contextmanager
def capture_logs():
    cap = LogCapture()
    old_processors = structlog.get_config()['processors']
    try:
        structlog.configure(processors=[cap])
        yield cap.entries
    finally:
        structlog.configure(processors=old_processors)
```

Yield value looks something like: `[{'error': ClientError('An error occurred (404) when calling the HeadObject operation: Not Found',), 'event': 'download_csv_error', 'log_level': 'critical'}]`

hynek:
@rgalanakis wow I love that approach! Simple yet powerful! I think I wouldn't be opposed to adding that to `structlog` proper!

rgalanakis:
Do you want a PR or will you add it yourself?

hynek:
I would like you to have the implementation credit but it will take a bit more than just the code. 🤔 I guess a Testing chapter is in order?

hynek:
[...]

PR #234, opened 2019-12-12
Add capture_logs context manager

rgalanakis:
See https://github.com/hynek/structlog/issues/14#issuecomment-563321043
for more context.

It needed to go into its own file, rather than `dev`,
because it relies on `_config` and `_config` relies on `dev`.


hynek:
I would suggest to create a module called `structlog.testing`. Maybe we'll find more testing helpers over the time.

rgalanakis:
Ok, renamed it. I also made LogCapture available for public use, and documented the pytest fixture use case.

hynek:
Thank you very much, I'll fix the version number myself . #procrastinatorExtraordinaire

PR #236, opened 2019-12-23
Fix #201 - Add support for contextvars

### LINKED ISSUE #201: Support for contextvars
Does structlog have support for contextvars similar to threadlocal?

They are new to Python 3.7, with a backport package available for Python 3.6 and 3.5, and IIUC correctly they are a reasonable replacement for threadlocal even when thread-local is the only requirement, but they also support async functions that would be run in the same thread.

http://www.structlog.org/en/stable/thread-local.html
https://docs.python.org/3/library/contextvars.html
https://github.com/MagicStack/contextvars

MarkusH:
This PR basically turns https://github.com/hynek/structlog/issues/201#issuecomment-557885793 into something mergeable.

Open question: How do I cleverly exclude the `test_contextvars.py` file on all Python 2 versions?


hynek:
Weee, what a xmas gift! ❤️

I'm not sure if there's a better way nowadays, but you can use [this approach](https://github.com/python-attrs/attrs/blob/b78720245a7944e5a091a075b7e9784fc93be05f/conftest.py#L16-L20) from attrs.

MarkusH:
> Weee, what a xmas gift! ❤️

Merry Christmas 🎄 🙂

> I'm not sure if there's a better way nowadays, but you can use [this approach](https://github.com/python-attrs/attrs/blob/b78720245a7944e5a091a075b7e9784fc93be05f/conftest.py#L16-L20) from attrs.

That's exactly what I was looking for, I think.

hynek:
Thanks, only fitting to merge it on Orthodox Christmas! ;)

MarkusH:
Thank you :heart: :sparkles:

WouldYouKindly:
This was the only thing that stopped me from trying out structlog. Thank you both.

PR #239, opened 2020-01-13
add unbind_threadlocal and try_unbind_threadlocal methods

c-wilson:
I really like the new processor-style interface for thread local context, but I was hoping we could extend it with unbind methods that mirror the `BoundLoggerBase` class. I frequently have the need to enter and exit a logging context, but I'm a bit worried about unintended side-effects using the old interface and `threadlocal.tmp_bind`.

adds:
* `structlog.threadlocal.unbind_threadlocal`
* tests

```python
import structlog

log= structlog.get_logger()

def myfunction():
  structlog.threadlocal.bind_threadlocal(x=12, y=2)
  log.msg('my message')
  # enter another function, module, etc.
  structlog.threadlocal.unbind_threadlocal('x', 'y')
```

If I'm violating the spirit of threadlocal context, just let me know. I would also be happy to add an actual logging context manager here or in another PR if it would be helpful!

```python
with LogContext(a=12, b=32):
  log.msg('mymsg')
```


MarkusH:
I'm not sure we need the strict and "lenient" implementations. Logging interactions should IMO not cause an application to fail. I'd thus go with a "fail-silent" approach as I've done in the contextvars implementation: https://github.com/hynek/structlog/blob/062aa214475f031596289170ea29e96e4157bd28/src/structlog/contextvars.py#L70-L72

hynek:
hmmm the problem is that it's not consistent with structlog's bind/unbind. I guess best way forward would be to add an option and deprecate try_undbind?

c-wilson:
Ha, now that I know about the contextvars implementation, I'm second-guessing all of this... 

While I do tend to agree that "logging should not make your app fail", there are definitely contexts where production of accurate logs are an important side-effect for guiding decisions (ie how much infra do I need, what is my ETL task failure rate, etc.).

Bringing to light double-removals of context could be valuable ie:

```python
def main(var)
  bind_threadlocal(var=var)
  fun1(var)
  fun2()
  log.msg("I finished something with 'a' in my context")
  unbind_threadlocal("var")

def fun1(var):
  bind_threadlocal(var=var)
  # do something
  unbind_threadlocal("var")

def fun2():
  pass
```

but maybe this highlights that we should use a context manager to roll-back context instead of this interface.

Anyways, let me know what you think, and I can remove try_unbind_threadlocal if that makes sense to you.

hynek:
Yeah let's just mirror the approach in contextvars please -- thanks!

hynek:
Thanks!

Could you add

- a change log entry
- and a `.. versionadded::` tag?

Thanks!

c-wilson:
Sure thing! Maybe this is a dumb question, but the versionadded will be `20.1.0`?

hynek:
Yeah but I'll do it myself real quick – thanks!

PR #242, opened 2020-01-25
Don't be lazy on Windows

### LINKED ISSUE #232: Default configuration does not look great for Windows users
It prints escape sequences to terminal.

Python 3.8.0
colorama v0.4.1
structlog v19.2.0

screenshot

(in foreground is powershell, background is cmd.exe, Windows 10)

hynek:
Fixes #232 

Could you check if this helps @wimglenn?

I've decided to go the grossest possible route: non-lazy on Windows, lazy everywhere else…

cc @MarkusH

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,7 @@ Changes:
   `#229 <https://github.com/hynek/structlog/issues/229>`_
 - ``colorama`` isn't initialized lazily on Windows anymore because it breaks rendering.
   `#232 <https://github.com/hynek/structlog/issues/232>`_,
+  `#242 <https://github.com/hynek/structlog/pull/242>`_
 
 
 ----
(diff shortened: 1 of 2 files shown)


MarkusH:
I don't have a Windows at hand right now. I can check the next days :)

wimglenn:
@hynek  I wasn't able to install your package on windows, something in the regex must be platform dependent

```
C:\...\structlog> python setup.py
  File "setup.py", line 100, in <module>
    + re.search(
AttributeError: 'NoneType' object has no attribute 'group'
```

However, I can confirm the fix works on Windows.  I was able to test this just by cd into the ./src directory and checking out the lazy-win-hack branch.

No shame in the "gross" route for windows - a dumb conditional is also how I deal with colorama init for cross-platform.

PR #245, opened 2020-02-08
Implement async bound loggers for stdlib

hynek:
This is so far only an experimental WiP/PoC. But unless I'm missing anything, adding truly async loggers shouldn't be much of a problem.

---

You can use it like this:

```python
import structlog
import asyncio
import logging
import sys

logging.basicConfig(
    format="%(message)s",
    stream=sys.stdout,
    level=logging.INFO,
)


structlog.configure(
    processors=[
        structlog.stdlib.filter_by_level,
        structlog.processors.UnicodeDecoder(),
        structlog.processors.JSONRenderer()
    ],
    context_class=dict,
    wrapper_class=structlog.stdlib.AsyncBoundLogger,  # <---
)

async def main():
    l = logger.bind(foo="bar")

    coro = l.info("qux")
    w = asyncio.sleep(0.5)

    await l.info("baz", x="42")

    await asyncio.gather(coro, w)

asyncio.run(main())
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
       exclude: docs/code_examples
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.3.0
+    rev: v3.4.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -53,7 +53,9 @@ Changes:
   `#266 <https://github.com/hynek/structlog/issues/266>`_,
 - ``structlog.PrintLogger`` now supports ``copy.deepcopy()``.
   `#268 <https://github.com/hynek/structlog/issues/268>`_
-- Added ``structlog.testing.CapturingLogger`` for more unittesting goodness.
+- Added ``structlog.testing.CapturingLogger`` for more unit testing goodness.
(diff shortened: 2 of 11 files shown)


Aeron:
I’m personally not sure how I feel about the thread executor approach. I understand that for such a rich library as Structlog—with various processors, handlers, and loggers—it might be even the only solution, to wrap a blocking logger calls (and everything behind it) and execute them in threads.

And yet I wonder if a simplistic log record emitting is worth the cost of thread creation and existence. I mean, the downsides of threads are well known. So it’s a compromise.

On the other hand, an Asyncio’s `StreamWriter` (more natural) approach seems to be much more resource-effective, but it implies everything in a chain must be non-blocking as well, of course.

Nonetheless, I’m glad to see it’s a WIP already, good job 👍

hynek:
> And yet I wonder if a simplistic log record emitting is worth the cost of thread creation and existence.

There are two problems here:

1. You cannot predict where most of the time will be spent. Processors am can take quite a bit of computing power (eg serialization).
2. Using async write APIs would require to rewrite all of structlog and all processors to be async too. I guess your best call would be to play with different loggers and I believe logging comes with something. But again: your logger will block while you json.dumps.

hynek:
@Aeron I've added a warning in a0d9d48209f6a361c0b4e26a4b9d2f6898cc9771 to point out the overhead. You're right that not everyone is gonna be aware of the trade-off.

PR #246, opened 2020-02-12
Implement add_function_name() processor in stdlib module

strokirk:
This PR adds a processor to add the `funcName` field that is automatically added when using standard library logging.

See: https://docs.python.org/3/library/logging.html#logrecord-attributes

# Pull Request Check List

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,7 +26,8 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-*none*
+- Added ``structlog.stdlib.add_function_name()`` processor that adds the name of function containing the logging call to the event dictionary.
+  `#246 <https://github.com/hynek/structlog/pull/246>`_
 
 
 ----
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -222,6 +222,8 @@ API Reference
 
 .. autofunction:: add_logger_name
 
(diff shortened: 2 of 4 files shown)


hynek:
This looks overall good but there's more subtleties to it, right? If I read it correctly, this processor would only work, if you pipe stdlib logs through structlog using the ProcessorFormatter?

strokirk:
Yes, exactly. You have to use the ProcessorFormatter to get the stdlib helpers to work, right?

Is there an easy way to capture the call stack for structlog-only logs?

hynek:
Sorry for my silence on this issue the reason is that there's a bunch of open issues around extracting information from log records. Most notably #253 which seems to be a super set of this problem? I guess this should be ideally integrated altogether.

strokirk:
Closing due to lack of interest, but it's fairly easy to just copy my code to any codebase that needs it!

hynek:
Yeah I'm sorry – I keep planning on a more complete solution and just can't find the strength/motivation. :(

strokirk:
Sending strength your way! ❤️ 
I think you might be right that there is a more general / all-compassing solution somewhere out there.

We've been moving away from structlog at our company due to the amount of setup needed to get it working (esp. with https://sentry.io), so no hard feelings from us. Still have an eye on this project, it's very interesting!

PR #247, opened 2020-02-18
Recommend another JSON library

ofek:
I've been using it for a few months now in a side project, where it beat out rapidjson & ujson in my basic benchmarks


hynek:
I think at this point (2020) we can remove the Python 3-only tags.

ofek:
Done!

hynek:
Thanks!

PR #257, opened 2020-04-23
Fix several typos

fbjorn:
Hi!

I was browsing the source code and noticed some typo in "iff" word (macbook keyboard?) 
But then I decided to go further and try out this nice [tool](https://github.com/vlajos/misspell-fixer), so found a bit more 🙂
There's also a [Github Action](https://github.com/sobolevn/misspell-fixer-action) for it if you think it's useful.

Thanks for the awesome library!

_contribution checklist is omitted, cause this is just about typos_


hynek:
Iff is short for "if and only if" which was a common abbreviation in my school science books? But since it seems that it keeps confusing people, I'll just remove it.

Thanks!

PR #263, opened 2020-06-16
Raise useful errors from PrintLogger with bad processor return

basepi:
I spent a good chunk of time fighting to get `PrintLogger` working with a simple processor. Turned out that processor was returning a dictionary, not a string. The [documentation](https://www.structlog.org/en/stable/processors.html) does point out that the final processor needs to provide a value that the logger can handle, but I assumed that a `PrintLogger` would just....print the value. (Python does print dicts, after all!)

Anyway, this PR doesn't propose changing that behavior, but rather adds a note to the docs, and also updates `PrintLogger` to take `*args` and `**kwargs`, throwing the `TypeError` itself with a more useful error message. I'm hoping that it will save future users new to structlog some valuable time.

# Pull Request Check List

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,6 +26,9 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
+- Added better error messages for PrintLogger when the processor chain doesn't return a string.
+  `#263 <https://github.com/hynek/structlog/pull/263>`_
+
 *none*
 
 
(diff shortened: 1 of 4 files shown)


hynek:
I'm sorry but this introduces a _runtime_ tax for a rather rare bug. I don't think this tradeoff is worth it.

basepi:
Would you accept a PR with just the docs changes? I think those are still useful.

basepi:
Opened a docs-only PR here: #269

PR #272, opened 2020-08-18
Fix the bug that ConsoleRenderer will throw exception when there is n…

youtian001:
The exception could be triggered if you rename the field _event_ in a preceding processor, for example from _event_ to _msg_.


hynek:
Thanks, hope that helps people running into this.

Issue #47, opened 2015-02-07
Document clean way to add custom log levels


heikoheiko:
Is there a different way, than accessing a private member like in below snippet?
```
structlog.stdlib.TRACE = TRACE = 5
structlog.stdlib._NAME_TO_LEVEL['trace'] = TRACE
logging.addLevelName(TRACE, "TRACE")
```

https://github.com/ethereum/pyethereum/issues/202

hynek:
Do you have any API or wishes in mind?

heikoheiko:
no not really. maybe something like .addLevelName( int_level, str_name)

hynek:
So I’m a bit ignorant about stdlib here, how does one add log levels there?

fxfitz:
stdlib logging levels are defined as integers (defaults are: https://docs.python.org/2/library/logging.html#logging-levels).

For example, `Logger.debug("OMG DEBUG MESSAGE")` would be equivalent to `Logger.msg(Logger.DEBUG, "OMG DEBUG MESSAGE")`

rydrman:
Has any more thought gone into this? It seems to me that a really good first step would be the ability for internal processes that use the logging level to resolve any integer to its closest recognized level. For example, say we have the standard recognized levels:
```
CRITICAL = 50
ERROR = 40
WARNING = 30
INFO = 20
DEBUG = 10
NOTSET = 0
```
in this case, logging at any level from 1-10 is simply understood as `DEBUG`, and anything over 50 is understood as `CRITICAL`

This would at least allow custom loggers to understand the defined levels without breaking anything in the structlog stdlib

pankajsharma1:
I tried above mentioned steps also
structlog.stdlib.TRACE = TRACE = 5
structlog.stdlib._NAME_TO_LEVEL['trace'] = TRACE
logging.addLevelName(TRACE, "TRACE")

But it is still not working for me with latest version

pdebelak:
@pankajsharma1 This is technically possible, but very difficult. See possible solution at https://stackoverflow.com/a/56467981/3945932.

alexbalandi:
Hello! Is the code provided by @darklow currently the best way to add custom log level or is there a cleaner API as of now?

hynek:
For console output, I think you can have it more elegantly with the new columns configuration: https://www.structlog.org/en/stable/console-output.html#console-output-configuration

[...]

PR #279, opened 2020-10-13
[WIP] [RFC] Add support for registering custom log levels

Kami:
This pull request adds some WIP code to get the discussion going on adding a new public API method for registering a custom log level with structlog.

## Background, Context

There have been many discussions and questions on the internet on how to register a custom log level with method (it's a fairly common request and problem). There are also open issues on that in this repo - e.g. #47.

I've used monkey patching / workaround similar to the ones proposed in that ticket elsewhere myself for many years now, but I think it would be good to expose this functionality directly in struclog and as such, make it less hacky, more official and less fragile.

## Proposed Implementation

This implementation is not finished yet.

It's just initial basic proposal to get the discussions going. Once we agree on the approach, I'm happy to fully implement it (with tests and other things of course).

### TODO

- [ ] Agree on the approach
- [ ] Implement it
- [ ] Add support for custom log levels to ConsoleRender

### DIFF
--- a/src/structlog/_base.py
+++ b/src/structlog/_base.py
@@ -205,3 +205,50 @@ class BoundLoggerBase:
             return getattr(self._logger, method_name)(*args, **kw)
         except DropEvent:
             return
+
+
+def register_log_level(level_name, level_value):
+    # type: (str, str) -> None
+    """
+    Register and add new custom log level with structlog and Python logging module.
+    """
+    import logging
+    import structlog.stdlib
+
+    level_name_lower = level_name.lower()
+    level_name_upper = level_name.upper()
+
+    # Check if the requested log level is already registered with stdlib
+    stdlib_log_level_value = getattr(logging, level_name.upper(), None)
+
+    if stdlib_log_level_value is not None:
+        raise ValueError("%s log level is already registered with stdlib logging and "
+                         "structlog" % (level_name.upper()))
+
+    # Check if the requested log level numeric value is already registed with stdlib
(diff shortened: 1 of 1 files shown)


hynek:
This belongs to `structlog.stdlib` but otherwise: is it code you use in prod like this?

I mean it's very gross 😅 but it is so due to `logging`'s design…I'd want to think over the addition of the logging methods but that's secondary.

I guess we should ask for feedback on #47?

hynek:
One idea before I forget it:

I think it would be better, to instead of globally monkeypatching stdlib.BoundLogger, to return a new stdlib.BoundLogger class.

src/structlog/_base.py
```diff
@@ -205,3 +205,50 @@ def _proxy_to_logger(self, method_name, event=None, **event_kw):
...
+        def log_method(self, msg, *args, **kwargs):
+            return self.log(level_value, msg, *args, **kwargs)
+        log_method.__name__ = name
+        return log_method
+
+    log_method = make_logger_function(level_name_lower)
+
+    setattr(structlog.stdlib._FixedFindCallerLogger, level_name_lower, log_method)
+    setattr(structlog.stdlib.BoundLogger, level_name_lower, log_method)
+
```
danilito19, line 252: I used some of this code for my project and I had to also add this line to get this to work
`setattr(logging.Logger, level_name_lower, log_method)`

hynek:
I think this is stale. I'm interested in a solution of this problem tho.

PR #324, opened 2021-06-01
Fix typo in performance.rst

jstvz:
Updated the sample code to include `TimeStamper(utc=True)` instead of `False`.

# Pull Request Check List

    - [X] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded).  Find the appropriate next version in our [``__init__.py``](https://github.com/hynek/structlog/blob/main/src/structlog/__init__.py) file.
- [X] Documentation in `.rst` files is written using [semantic newlines](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).
- [X] Changes (and possible deprecations) are documented in the [changelog](https://github.com/hynek/structlog/blob/main/CHANGELOG.rst).

### DIFF
--- a/docs/performance.rst
+++ b/docs/performance.rst
@@ -64,7 +64,7 @@ Here's an example for a production-ready non-asyncio ``structlog`` configuration
           structlog.threadlocal.merge_threadlocal_context,
           structlog.processors.add_log_level,
           structlog.processors.format_exc_info,
-          structlog.processors.TimeStamper(fmt="iso", utc=False),
+          structlog.processors.TimeStamper(fmt="iso", utc=True),
           structlog.processors.JSONRenderer(serializer=orjson.dumps),
       ],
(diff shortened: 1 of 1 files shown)


hynek:
Would you mind elaborating why you consider it a typo?

jstvz:
> Would you mind elaborating why you consider it a typo?

Of course. Below it states that:

> Adds an `ISO 8601 <https://en.wikipedia.org/wiki/ISO_8601>`_ timestamp under the ``timestamp`` key in the UTC timezone.

My understanding is that passing `utc=False` would result in local time being used.

hynek:
Ha, you're right – thanks!

Those stupid templates/check lists discourage people from writing their own thoughts. I really wish there were a better way. :/

PR #335, opened 2021-07-14
add processor_chain arg to stdlib.ProcessorFormatter

### LINKED ISSUE #334: multiple processors to structlog.stdlib.ProcessorFormatter
I want structlog messages and stdlib logging to go to stdout and json file. I want timestamps in json file, but not in stdout.

Currently I have to do this (based on [this part of the docs](https://www.structlog.org/en/stable/standard-library.html#rendering-using-structlog-based-formatters-within-logging)):

```python
add_timestamp = structlog.processors.TimeStamper(fmt="%Y-%m-%d %H:%M:%S")
render_json = structlog.processors.JSONRenderer()

def add_timestamp_and_render_json(logger, name, event_dict):
    event_dict = add_timestamp(logger, name, event_dict)
    event_dict = render_json(logger, name, event_dict)
    return event_dict

logging.config.dictConfig({
    ...
    "formatters": {
        "json_formatter": {
            "foreign_pre_chain": [
            ],
            "processor": add_timestamp_and_render_json,
        },
        "stdout_formatter": {
            "foreign_pre_chain": [
            ],
            "processor": structlog.dev.ConsoleRenderer(),
        },
    },
    ...
})
```

Would be nice if I could provide a chain instead of just one `processor`:

```python
        "json_formatter": {
            ...
            "processor_chain": [
                structlog.processors.TimeStamper(fmt="%Y-%m-%d %H:%M:%S"),
            ]
        },
```

Akuli:
Fixes #334 

I'll add tests if any of this is actually a good idea. I'm new to structlog and there might already be a way to do what I wanted to do in #334 .

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -32,6 +32,7 @@ Changes:
 - If the `better-exceptions <https://github.com/qix-/better-exceptions>`_ package is present, ``structlog.dev.ConsoleRenderer`` will now pretty-print exceptions using it.
   Pass ``pretty_exceptions=False`` to disable.
   This only works if ``format_exc_info`` is **absent** in the processor chain.
+- Added ``processor_chain`` argument to ``structlog.stdlib.ProcessorFormatter``.
 
 
 ----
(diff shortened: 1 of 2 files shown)


hynek:
I'm sorry, but we can't just remove such an important parameter.

Shouldn't it be possible to write a processor that wraps a chain? Processors should be composable, so maybe we should add more helpers to achieve that.

Akuli:
That makes sense. This just isn't a good approach.

hynek:
Hi @Akuli, ironically I'm working on something very similar. Would #365 work for you?

Akuli:
Thanks! Seems like it would work, although I no longer work on the project where I wanted this.

PR #337, opened 2021-07-17
Add get_contextvars & get_merged_contextvars

hynek:
Finalize #327


twosigmajab:
Thanks for adding this! Coincidentally, I just hit a case that needs this functionality. Would it be an okay time to cut a new release sometime soon?

hynek:
I would like to tackle stdlib metadata first but don’t know when I’ll find the time. Why don’t you just copy paste it for now?

PR #339, opened 2021-07-26
Add reset_contextvars and update bind_contextvars

jab:
Update `structlog.contextvars.bind_contextvars()` to return a `Mapping[str, contextvar.Token]` corresponding to the results
of setting the requested contextvars.

This return value is suitable for use with the newly-added `structlog.contextvars.reset_contextvars()` function, which resets the requested contextvars to their previous values using the given Tokens.

# Pull Request Check List

### DIFF
--- a/docs/contextvars.rst
+++ b/docs/contextvars.rst
@@ -47,6 +47,7 @@ The general flow is:
    >>> # values:
    >>> clear_contextvars()
    >>> bind_contextvars(a=1, b=2)
+   {'a': <Token var=<ContextVar name='structlog_a' default=Ellipsis at ...> at ...>, 'b': <Token var=<ContextVar name='structlog_b' default=Ellipsis at ...> at ...>}
    >>> # Then use loggers as per normal
    >>> # (perhaps by using structlog.get_logger() to create them).
    >>> log.msg("hello")
@@ -61,3 +62,18 @@ The general flow is:
    >>> clear_contextvars()
    >>> log.msg("hi there")
    event='hi there' a=None
+
+
+If e.g. your request handler calls a helper function that needs to temporarily override some contextvars before restoring them back to their original values, you can use the :class:`~contextvars.contextvars.Token`\s returned by :func:`~structlog.contextvars.bind_contextvars` along with :func:`~structlog.contextvars.reset_contextvars` to accomplish this (much like how :meth:`contextvars.ContextVar.reset` works):
+
+.. code-block:: python
+
+    def foo():
+        bind_contextvars(a=1)
+        _helper()
+        log.msg("a is restored!")  # a=1
+
+    def _helper():
+        tokens = bind_contextvars(a=2)
+        log.msg("a is overridden")  # a=2
+        reset_contextvars(**tokens)
--- a/src/structlog/contextvars.py
+++ b/src/structlog/contextvars.py
@@ -16,7 +16,7 @@ See :doc:`contextvars`.
 
 import contextvars
 
-from typing import Any, Dict
+from typing import Any, Dict, Mapping
 
 import structlog
 
@@ -98,16 +98,22 @@ def clear_contextvars() -> None:
             k.set(Ellipsis)
 
 
-def bind_contextvars(**kw: Any) -> None:
-    """
+def bind_contextvars(**kw: Any) -> "Mapping[str, contextvars.Token[Any]]":
+    r"""
     Put keys and values into the context-local context.
 
     Use this instead of :func:`~structlog.BoundLogger.bind` when you want some
     context to be global (context-local).
 
+    Return the mapping of ``contextvars.Token``\s resulting
+    from setting the backing ``ContextVar``\s.
+    Suitable for passing to :func:`reset_contextvars`.
+
     .. versionadded:: 20.1.0
-    .. versionchanged:: 21.1.0 See toplevel note.
+    .. versionchanged:: 21.1.0 Return the ``contextvars.Token`` mapping
+        rather than None. See also the toplevel note.
     """
+    rv = {}
     for k, v in kw.items():
         structlog_k = f"{STRUCTLOG_KEY_PREFIX}{k}"
         try:
@@ -116,7 +122,21 @@ def bind_contextvars(**kw: Any) -> None:
             var = contextvars.ContextVar(structlog_k, default=Ellipsis)
             _CONTEXT_VARS[structlog_k] = var
 
-        var.set(v)
+        rv[k] = var.set(v)
+
+    return rv
+
+
+def reset_contextvars(**kw: "contextvars.Token[Any]") -> None:
+    r"""
+    Reset contextvars corresponding to the given Tokens.
+
+    .. versionadded:: 21.1.0
+    """
+    for k, v in kw.items():
+        structlog_k = f"{STRUCTLOG_KEY_PREFIX}{k}"
+        var = _CONTEXT_VARS[structlog_k]
+        var.reset(v)
 
 
 def unbind_contextvars(*keys: str) -> None:
--- a/tests/test_contextvars.py
+++ b/tests/test_contextvars.py
@@ -16,6 +16,7 @@ from structlog.contextvars import (
     get_contextvars,
     get_merged_contextvars,
     merge_contextvars,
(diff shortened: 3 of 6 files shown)


hynek:
I’m gonna have to ask you to write some narrative docs for contextvars.rst so I have a clue what this is about. 😅 

i have to admit that my understanding of contextvars is rather superficial.

jab:
Done in the latest revision. Is it any clearer now?

src/structlog/contextvars.py
```diff
@@ -98,16 +98,22 @@ def clear_contextvars() -> None:
             k.set(Ellipsis)
 
 
-def bind_contextvars(**kw: Any) -> None:
-    """
+def bind_contextvars(**kw: Any) -> Mapping[str, contextvars.Token]:
```
jab, line 101: I think for some reason, Sphinx is getting confused trying to resolve the occurrences of `contextvar.Token` in these typehints and docstrings, causing a spurious warning (which is getting turned into a spurious error).

I noticed the [old-(pre-Sphinx-1.0)-format `intersphinx_mapping`](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping) was being used, so I tried updating it to the up-to-date format, but that wasn't enough to appease Sphinx. 🤷

jab: Meant to say, I also tried several other things (e.g. adding `from contextvars import Token` and changing the typehint to just `Token`, and when that didn't work, tried spelling these references a bunch of other ways), but nothing I tried worked, and I gave up after the better part of an hour. I wonder if it's confusing Sphinx that both `structlog` and the standard library have modules named `contextvars`?

hynek: Unfortunately Sphinx and types is hit and miss. :(

If you can't make it work, add it to https://github.com/hynek/structlog/blob/b39a720df2a8b144ecafd9d7eedaba9d0ecdf5dc/docs/conf.py#L69-L81

docs/contextvars.rst
```diff
@@ -61,3 +62,23 @@ The general flow is:
...
    >>> log.msg("hi there")
    event='hi there' a=None
+
+
+If e.g. your request handler calls a helper function that needs to
+temporarily override some contextvars before restoring them back to
+their original values, you can use the :class:`~contextvars.contextvars.Token`\s
+returned by :func:`~structlog.contextvars.bind_contextvars` along with
+:func:`~structlog.contextvars.reset_contextvars` to accomplish this
+(much like how :meth:`contextvars.ContextVar.reset` works):
```
hynek, line 72:
```suggestion
If e.g. your request handler calls a helper function that needs to temporarily override some contextvars before restoring them back to their original values, you can use the :class:`~contextvars.contextvars.Token`\s returned by :func:`~structlog.contextvars.bind_contextvars` along with :func:`~structlog.contextvars.reset_contextvars` to accomplish this (much like how :meth:`contextvars.ContextVar.reset` works):
```
We use [semantic newlines](http://rhodesmill.org/brandon/2012/one-sentence-per-line/).

src/structlog/contextvars.py
```diff
@@ -116,7 +122,20 @@ def bind_contextvars(**kw: Any) -> None:
             var = contextvars.ContextVar(structlog_k, default=Ellipsis)
             _CONTEXT_VARS[structlog_k] = var
 
-        var.set(v)
+        rv[k] = var.set(v)
+    return rv
```
hynek, line 126:
```suggestion

    return rv
```

tests/test_contextvars.py
```diff
@@ -63,6 +64,25 @@ async def coro():
...
+    async def test_reset(self, event_loop):
+        """
+        reset_contextvars allows resetting contexvars to
+        previously-set values.
+        """
+
+        async def coro():
+            bind_contextvars(a=1)
+            assert {"a": 1} == get_contextvars()
+            await event_loop.create_task(nested_coro())
```
hynek, line 76:
```suggestion
            bind_contextvars(a=1)
            
            assert {"a": 1} == get_contextvars()
            
            await event_loop.create_task(nested_coro())
```

tests/test_contextvars.py
```diff
@@ -63,6 +64,25 @@ async def coro():
...
+
+        async def nested_coro():
+            tokens = bind_contextvars(a=2, b=3)
+            assert {"a": 2, "b": 3} == get_contextvars()
+            reset_contextvars(**tokens)
+            assert {"a": 1} == get_contextvars()
```
hynek, line 82:
```suggestion
            assert {"a": 2, "b": 3} == get_contextvars()
            
            reset_contextvars(**tokens)
            
            assert {"a": 1} == get_contextvars()
```

hynek (changes requested):
Just a few stylistic changes.

Other than that, it needs a changelog entry. I would suggest some high-level description telling why ppl should be excited?

Maybe something like:

```rst
``structlog.contextvars.bind_contextvars()`` now returns a mapping of keys to ``contextvars.Token``\ s that allow to reset values using the new ``structlog.contextvars.reset_contextvars()``.
```

docs/contextvars.rst
jab: Done in the latest revision.

(Though, as a fan of semantic line breaks myself, I was surprised to get this suggestion. I think adding additional line breaks at appropriate positions within a single sentence is actually encouraged, not forbidden. See the example at https://sembr.org starting with "We can further clarify the source text by adding a line break after the clause". This makes reading and editing docs (and diffs) easier in editors that are not set to soft-wrap long lines, as well as pagers and other tools.)

src/structlog/contextvars.py
jab: Thanks! Done in the latest revision. (After rebasing on top of latest `main`, I also had to add an entry for bare `"TLLogger"`, to ignore a new warning I was getting that is unrelated to my changes.)

src/structlog/contextvars.py
jab: Done in the latest revision.

tests/test_contextvars.py
jab: Done in the latest revision.

tests/test_contextvars.py
jab: Done in the latest revision.

jab (commented):
Thanks for reviewing, @hynek! I believe I've addressed all your requested changes.

hynek (changes requested):
mypy is failing:

```
src/structlog/contextvars.py:101: error: Missing type parameters for generic type "Token"
src/structlog/contextvars.py:130: error: Missing type parameters for generic type "Token"
```

I suspect (vacationing rn) you have to set it to `Token[Any]`?

jab:
@hynek should be fixed now! Thanks for catching that, and sorry I missed that when testing locally. (Each time I update this PR, the CI checks don't run until you approve running them, since this is my first structlog PR. If there were some way for you to pre-approve running the checks for future updates of this PR, I would have noticed and fixed that by the time you took your next look. Not sure if that's possible?)

Hope you've been having a great vacation!

hynek:
Sorry terse because phone:

- any reason why running `tox` Locally doesn’t work for you?
- I think we’ll have to go for a string annotation :(
- I’m not aware of a way to pre-approve users. If you find anything lmk. But it’ll be moot once a single PR by you is merged.

jab:
`tox` is now succeeding for me locally as of the latest revision. I think 3.6, 3.7, 3.8, 3.9, and docs should all succeed in CI once they run next.

Not sure what's going on with the required codecov status checks. Please let me know if there's anything else you need me to do about those. Thanks for reviewing!

hynek:
Quickly approved before leaving. As long as builders are failing, you can ignore codecov, because of fast-failing alone (you can see that some builders don’t even start).

jab:
> All checks have passed

🎉

hynek:
Thanks for your perseverance!

jab:
Thanks for merging! And for structlog and your other great work!

hynek:
JFTR, I've run into `contextvars.Token` problems myself and it's actually a bug in CPython's docs!

So I've fixed it today: https://github.com/python/cpython/pull/29533

PR #345, opened 2021-08-06
Cease all use of Colorama on non-Windows systems.

coredumperror:
    - [X] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded).  Find the appropriate next version in our [``__init__.py``](https://github.com/hynek/structlog/blob/main/src/structlog/__init__.py) file.
- [X] Documentation in `.rst` files is written using [semantic newlines](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).
- [X] Changes (and possible deprecations) are documented in the [changelog](https://github.com/hynek/structlog/blob/main/CHANGELOG.rst).

This PR was inspired by the frustration and confusion I suffered over the last week, dealing with figuring out why colored logging wan't working on my apps, despite everything seemingly having been configured properly. I eventually tracked it down to the `self.stream.isatty()` check on `colorama`'s `ansitowin32.py` file, line 92. Because my apps run in Docker containers, they don't log directly to a TTY, so it was returning False and forcing the ANSI-to-Win32 color code conversion, even though my apps run on Linux. This broke all the colors.

Yes, fixing this is what `ConsoleRenderer`'s `force_colors` kwarg appears to be for, but I was using an old version of `ConsoleRenderer` from before that was added, and didn't realize it.

Ultimately, though, my experience showed me that `colorama` isn't intended to be used on non-Windows systems at all (See https://github.com/tartley/colorama/issues/306). The *only* thing `colorama.init()` does is wrap `sys.stdout` and `sys.stderr` in the `AnsiToWin32` wrapper, which either does nothing on non-Win32 systems (if logging to a TTY) or is actively harmful (if logging to a file).

So I believe that `structlog` should change to avoid the use of `colorama` *at all* unless it's running on Windows. That's what this PR does. I made the following changes:

1. Replaced the blank color codes that `structlog` defines if `colorama` isn't installed with the actual raw ANSI color codes that `colorama` generates when one uses those constants.
2. Removed the lazy init code for `colorama`, since it isn't done lazily on Windows, and shouldn't be done on other OSes at all.
3. Fixed what appears to be a bug with the `force_colors` kwarg on Windows. The Windows init code was using `self._force_colors`, which was always `False` where that code was being called, even if the `force_colors` kwarg was `True`.
4. Changed the default value of `ConsoleRenderer`'s `colors` kwarg from "true if `colorama` is installed" to "True on Windows if `colorama` is installed, but always True on other OSes".
5. Updated the docs, tests, and changelog to reflect this change.

I hope I did this right... I don't contribute to GitHub all that often, but your "How to Contribute" guide was freaking **great** for helping give me the confidence that I was at least mostly doing this right. I ran `tox` and `pre-commit` for the first time ever, and it showed me some bugs in my original code, which I fixed before committing.

The one thing I'm a bit iffy on is the tests. I really just updated the "Make sure this crashes if `colors=True` is sent when colorama isn't installed" test doesn't get run unless it's running on Windows. I'm not really certain that's all that needs to change, or if that's even the right change.

### DIFF
--- a/src/structlog/dev.py
+++ b/src/structlog/dev.py
@@ -47,8 +47,6 @@ def _pad(s: str, length: int) -> str:
 
 
 if colorama is not None:
-    _has_colorama = True
-
     RESET_ALL = colorama.Style.RESET_ALL
     BRIGHT = colorama.Style.BRIGHT
     DIM = colorama.Style.DIM
@@ -60,21 +58,27 @@ if colorama is not None:
     GREEN = colorama.Fore.GREEN
     RED_BACK = colorama.Back.RED
 else:
-    _has_colorama = False
-
     # These are the same values as the colorama color codes. Redefining them
-    # here allows users to specify that they want color without having to 
+    # here allows users to specify that they want color without having to
     # install colorama, which is only supposed to be necessary in Windows.
-    RESET_ALL = '\033[0m'
-    BRIGHT = '\033[1m'
-    DIM = '\033[2m'
-    RED = '\033[31m'
-    BLUE = '\033[34m'
-    CYAN = '\033[36m'
-    MAGENTA = '\033[35m'
-    YELLOW = '\033[33m'
-    GREEN = '\033[32m'
-    RED_BACK = '\033[41m'
+    RESET_ALL = "\033[0m"
+    BRIGHT = "\033[1m"
+    DIM = "\033[2m"
+    RED = "\033[31m"
+    BLUE = "\033[34m"
+    CYAN = "\033[36m"
+    MAGENTA = "\033[35m"
+    YELLOW = "\033[33m"
+    GREEN = "\033[32m"
+    RED_BACK = "\033[41m"
+
+
+if _IS_WINDOWS:
+    # On Windows, use colors by default only if colorama is installed.
+    _use_colors = colorama is not None
+else:
+    # On other OSes, use colors by default.
+    _use_colors = True
 
 
 class _Styles(Protocol):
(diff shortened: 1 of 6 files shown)


src/structlog/dev.py
```diff
@@ -174,34 +188,36 @@ class ConsoleRenderer:
...
+                    raise SystemError(
+                        _MISSING.format(
+                            who=classname + " with `colors=True`",
+                            package="colorama",
+                        )
+                    )
+                else:
+                    # Colorama must be init'd on Windows, but must NOT be
+                    # init'd on other OSes, because it can break colors.
+                    _init_colorama(force_colors)
```
hynek, line 220: Since the `if` ends with a `raise`, drop the `else`. Then  inline `_init_colorama` since – as far as I can see – it's now only used here.. That should fix your coverage problems (the whole branch is tagged as `no cover`).

hynek (changes requested):
I'm glad all the typing paid off in the end! :)

You've almost got it, it's just the semantic new lines that got lost in the litany.

I have left instructions that should fix the coverage problems too.

Getting rid of a dependency on most platforms sounds GREAT!

coredumperror:
OK, I've made the changes you suggested, and pushed them to this PR.

src/structlog/dev.py
```diff
@@ -60,11 +58,27 @@ def _pad(s: str, length: int) -> str:
...
+    RED = "\033[31m"
+    BLUE = "\033[34m"
+    CYAN = "\033[36m"
+    MAGENTA = "\033[35m"
+    YELLOW = "\033[33m"
+    GREEN = "\033[32m"
+    RED_BACK = "\033[41m"
+
+
+if _IS_WINDOWS:
```
hynek, line 76:
```suggestion
if _IS_WINDOWS:  # pragma: no cover
```

hynek (changes requested):
ah sorry forgot about this one. once it's applied, we should be ready to merge!

hynek (approved):

hynek:
Thanks!

PR #364, opened 2021-11-08
Get TimeStamper working properly if datetime module is mocked before its instantiation

moriyoshi:
## Summary

Because the reference to either `datetime.now()` or `datetime.utcnow()` is retrieved during `TimeStamper` is being created, it gets into trouble when FreezeGun or any similar mechanism that patches `datetime` module is applied before the instantiation.  This patchs proposes a modification to it so that the invocation to the "now" function will always be done with dereferences from the datetime module object, albeit the slight possible performance penalty.

## Checklist

### DIFF
--- a/src/structlog/processors.py
+++ b/src/structlog/processors.py
@@ -365,11 +365,11 @@ def _make_stamper(
 
         if utc:
             return stamper_iso_utc
-        else:
-            return stamper_iso_local
+
+        return stamper_iso_local
 
     def stamper_fmt(event_dict: EventDict) -> EventDict:
-        event_dict[key] = now().strftime(fmt)
+        event_dict[key] = now().strftime(fmt)  # type: ignore
 
         return event_dict
 
--- a/src/structlog/processors.py
+++ b/src/structlog/processors.py
@@ -333,7 +333,17 @@ def _make_stamper(
     if fmt is None and not utc:
         raise ValueError("UNIX timestamps are always UTC.")
 
-    now = getattr(datetime.datetime, "utcnow" if utc else "now")
+    now: Callable[[], datetime.datetime]
+
+    if utc:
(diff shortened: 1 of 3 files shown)


src/structlog/processors.py
```diff
@@ -333,7 +333,17 @@ def _make_stamper(
     if fmt is None and not utc:
         raise ValueError("UNIX timestamps are always UTC.")
 
-    now = getattr(datetime.datetime, "utcnow" if utc else "now")
+    now: Callable[[], datetime.datetime]
```
moriyoshi, line 336: * This declaration is not necessary, but I prefer the "Φ" to be obvious.

hynek (approved):
I guess it's a slight performance regression but function calls are getting cheaper over the past releases. :)

hynek (changes requested):
Looks like mypy is not happy:

```
src/structlog/processors.py:372: error: Argument 1 to "strftime" of "datetime" has incompatible type "Optional[str]"; expected "str"
```

moriyoshi:
> Looks like mypy is not happy:
> src/structlog/processors.py:372: error: Argument 1 to "strftime" of "datetime" has incompatible type "Optional[str]"; expected "str"

This looks like a false error because `fmt` should have been asserted as non-None by the preceding `if` block and I didn't touch the part.

hynek:
> > Looks like mypy is not happy:
> > src/structlog/processors.py:372: error: Argument 1 to "strftime" of "datetime" has incompatible type "Optional[str]"; expected "str"
> 
> This looks like a false error because `fmt` should have been asserted as non-None by the preceding `if` block and I didn't touch the part.

You're right, looks like a mypy bug!

hynek (approved):

Issue #218, opened 2019-06-25
Is better context manager possible?


slavashvets:
Hello everyone. I'm looking for the better way how to add context values to log messages. The existing `tmp_bind` doesn't look perfect for me because it forces me to pass logger as an argument to another function.

After some experimentation, I got the following:

```python
class bind():
  def __init__(self, *args, **kwargs):
    self.kwargs = kwargs

  def __enter__(self):
    global logger
    logger = logger.bind(**self.kwargs)
  
  def __exit__(self, exc_type, exc_val, exc_tb):
    global logger
    logger = logger.unbind(*self.kwargs.keys())

def create_bucket(bucket: str):
  with bind(bucket=bucket):
    logger.info("Creating bucket")
    minio.make_bucket(bucket)

def create_buckets(buckets: list):
  with bind(action='creating'):
    for bucket in buckets:
      create_bucket(bucket)
    logger.info("Completed for all buckets")

create_buckets(['test-private'])
```
In the logs I see:
```
[info     ] Creating bucket             action=creating bucket=test-private
[info     ] Completed for all buckets   action=creating
```
Exactly what I need.
However, I don’t like to use "global" keyword. Do you know a better way?

hynek:
I assume you're using threadlocals here? You don't need the logger variable for writing, so you should be able to just make the functiin bodies `logger.bind()` and `logger.unbind()`?

hynek:
BTW, I know it's not exactly what you wanted, but depending on your use case #225 might be interesting to you nevertheless!

fernandobrito:
This page is the first Google result for "structlog context manager", so I'm sharing my approach here:

```python
@contextlib.contextmanager
def bind_on_logger(**kwargs):
    structlog.threadlocal.bind_threadlocal(**kwargs)
    try:
        yield
    finally:
        structlog.threadlocal.unbind_threadlocal(*kwargs.keys())
```

And then I can just do:
```python
logger = get_logger()

logger.info("outer")

with bind_on_logger(a=1):
    logger.info("inner 1")

    with bind_on_logger(b=2):
        logger.info("inner 2")

# Also available as a decorator

@bind_on_logger(c=3)
def my_func():
    logger.info("function call")


my_func()
```

to get:

![image](https://user-images.githubusercontent.com/196840/140549753-e2b24518-3d47-4734-9fa9-249f87a7ad36.png)

fernandobrito:
[...]

PR #371, opened 2021-11-21
Add threadlocal bound context manager

fernandobrito:
Adds a context manager/decorator similar to `tmp_bind`, but using the thread-local context instead.

Follow-up from https://github.com/hynek/structlog/issues/218.

I have some questions before I start with documentation and improve the tests.
I'm also new to `structlog`, so let me know if I'm missing something obvious.

By the way, great PR checklist, contributing guide and setup with `pre-commit` 🙌 .

## Clean up

When exiting the context manager, how should we handle unbiding?

A) We could just `unbind_threadlocal` all keys that were passed as parameters to the context manager (like this version of the PR), but if there was already an existing key in the context before the use of `bound` and `bound` overwrites it with another value, in this approach the teardown step will actually remove whatever value was there before. See sample test `test_cleanup_conflict` in this PR.

B) `tmp_bind` creates a copy of the context and "rewinds" it when tearing it down. If we use this approach here, we will also revert any other independent use of `bind_threadlocal` or `unbind_threadlocal` that happened inside the managed context. See sample test `test_cleanup_conflict2` in this PR. Personally, I prefer the approach above since it seems less likely to affect users than this one.

What are your thoughts on it? Keep it simple, use approach A and make it clear in the documentation that conflicting keys will be removed after the context manager is done? Or go nuts and try an option C where we would check what keys `bound` is overwriting, then we keep only a copy of those (and not of the whole context, like approach B), and on teardown we first `unbind` all keys passed to `bound` and then merge this copy back to the context? Is it worth complicating it like that?

Let me know if this all makes sense or if I should try to rephrase my thoughts 😁

# Pull Request Check List


hynek:
I think as a user, I'd expect my old key to be restored and nothing else.

IOW:

```python
threadlocal_bind(x=1, y=10)
with threadlocal_bound(x=2):  # please change to this name :)
    log.info("foo") # logs x=2, y=10

 log.info("foo") # logs x=1, y=10
```

Your nuclear option would be more of a snapshot feature than a temporary bind. 😅

---

I guess you'll have to save which keys you need to delete and which you need to restore before binding and then do it in finally.

fernandobrito:
Great, thanks for taking a look! I agree with you, and I will address it when time allows (probably only in the next weekend).

hynek:
Cool! Since we've had a minor regression in #373, I'd like to push a new release sometime next week. Would be great if we could ship this along with it!

docs/api.rst
```diff
@@ -119,6 +119,10 @@ Modern Approach
 
 .. autofunction:: bind_threadlocal
 
+.. autofunction:: unbind_threadlocal
```
fernandobrito, line 122: `unbind_threadlocal` was also missing from here.

fernandobrito:
@hynek AwesomeI just did some small changes and improved the overwriting behavior/restoring process.

Regarding the documentation for the end-users, I'm not sure where to add it.
Currently, `tmp_bind` is actually inside the `Wrapped Dicts` section of `thread-local`: https://www.structlog.org/en/stable/thread-local.html#wrapped-dicts. Should I move `tmp_bind` to a new section and add `bound_threadlocal` close to it, or what would you suggest?

CHANGELOG.rst
```diff
@@ -26,6 +26,9 @@ Changes:
 
 - Fixed import when running in optimized mode (``PYTHONOPTIMIZE=2``).
   `#373 <https://github.com/hynek/structlog/pull/373>`_
+- Added ``structlog.threadlocal.bound_threadlocal`` as a thread-local equivalent of ``structlog.threadlocal.tmp_bind``
+  for temporarily binding keys to the global context through a decorator/context manager.
```
hynek, line 30: Let's not "advertise" the old APIs. Also: semantic newlines please. :)

```suggestion
- Added the ``structlog.threadlocal.bound_threadlocal`` decorator/context manager to temporarily bind key/value pairs to a thread-local context.
```

src/structlog/threadlocal.py
```diff
@@ -93,6 +93,9 @@ def tmp_bind(
 ) -> Generator[TLLogger, None, None]:
     """
     Bind *tmp_values* to *logger* & memorize current state. Rewind afterwards.
+
+    Use :func:`~structlog.threadlocal.bound_threadlocal` instead when you want
+    the temporary bindings to be applied to the global context (thread-local).
```
hynek, line 98:
```suggestion
    Only works with `structlog.threadlocal.wrap_dict`-based contexts.
    Use :func:`~structlog.threadlocal.bound_threadlocal` for new code.
```

src/structlog/threadlocal.py
```diff
@@ -255,6 +258,28 @@ def unbind_threadlocal(*keys: str) -> None:
...
 
 
+@contextlib.contextmanager
+def bound_threadlocal(**kw: Any) -> None:
+    """
+    Bind *kw* to the current thread-local context. Unbind *kw* afterwards &
+    restore overwritten keys. Can be used as a context manager or decorator.
+
+    Use this instead of :func:`~structlog.threadlocal.tmp_bind` when you want
+    the temporary bindings to be applied to the global context (thread-local).
```
hynek, line 268: Let's not point to `tmp_bind`, the new APIs stand for themselves:

```suggestion
    Bind *kw* to the current thread-local context. Unbind or restore *kw*
    afterwards. Do **not** affect other keys.

    Can be used as a context manager or decorator.
```

hynek (changes requested):
Looking good!

Just please:

- add an example to tests/typing_examples.py
- add an example to the big example box in thread-local.rst, line 46pp

hynek:
> Regarding the documentation for the end-users, I'm not sure where to add it. Currently, `tmp_bind` is actually inside the `Wrapped Dicts` section of `thread-local`: https://www.structlog.org/en/stable/thread-local.html#wrapped-dicts. Should I move `tmp_bind` to a new section and add `bound_threadlocal` close to it, or what would you suggest?

No, please leave `tmp_bind` alone, I want people to forget about it. 😅

Just add it to the big doctest in line 46 et pp. :)

CHANGELOG.rst
```diff
@@ -26,6 +26,9 @@ Changes:
...
+ `#371 <https://github.com/hynek/structlog/pull/371>`_
```
hynek, line 31:
```suggestion
- Added the ``structlog.threadlocal.bound_threadlocal`` decorator/context manager to temporarily bind key/value pairs to a thread-local context.
  `#371 <https://github.com/hynek/structlog/pull/371>`_
```

hynek:
Looks like the types need some love: https://github.com/hynek/structlog/runs/4320516792?check_suite_focus=true#step:5:17

Is that something you're comfortable to tackle?

fernandobrito:
> Looks like the types need some love: https://github.com/hynek/structlog/runs/4320516792?check_suite_focus=true#step:5:17
> 
> Is that something you're comfortable to tackle?

Yes, I think so, especially because there are other examples in the code already and `mypy` is already pointing me somewhere, but since tonight I will be busy, I will only be able to work on it tomorrow evening.

hynek:
I think I've got it – thanks!

fernandobrito:
Awesome, thanks! 🙌

PR #377, opened 2021-12-05
Add `ExtraAdder` Processor function to `structlog.stdlib`

### LINKED ISSUE #209: Log stdlib's logging `extra`-s
Quite many people use python's stdlib logging with extras, say, `logger.info("message", extra={'a': 'b'})`. It's quite obvious how to format these messages in a customer way with stdlib, and many formatters (say, `LogstashV1Formatter`) render them by default. However, these are ignored when the rest of the logging and rendering is done via structlog. 

Would be great to have the extra dictionary contents from stdlib logging passed to structlog, so
```
stdlib_logger.info("TEST from %s logger", 'stdlib', extra={'a': 'b'})
```
and
```
struct_logger.info("TEST from %s logger", 'struct', a='b')
```
would produce equivalent results.

Typical use-case:
 * the underlying library uses stdlib logging,
 * the application uses structlog.

aucampia:
# Summary

Added the `structlog.stdlib.ExtraAdder` processor that adds extra attributes of `logging.LogRecord` objects to the event dictionary.
  This processor can be used for adding data passed in the `extra` parameter of the `logging` module's log methods to the event dictionary.

Fixes #209

# Pull Request Check List

This is just a friendly reminder about the most common mistakes.  Please make sure that you tick all boxes.  But please read our [contribution guide](https://github.com/hynek/structlog/blob/main/.github/CONTRIBUTING.md) at least once, it will save you unnecessary review cycles!

- [x] **New APIs** are added to [`typing_examples.py`](https://github.com/hynek/structlog/blob/main/tests/.py).
    - [x] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded).  Find the appropriate next version in our [`__init__.py`](https://github.com/hynek/structlog/blob/main/src/structlog/__init__.py) file.

### DIFF
--- a/src/structlog/stdlib.py
+++ b/src/structlog/stdlib.py
@@ -18,6 +18,7 @@ from functools import partial
 from typing import (
     Any,
     Callable,
+    Collection,
     Dict,
     Iterable,
     List,
@@ -44,7 +45,7 @@ __all__ = [
     "add_log_level_number",
     "add_log_level",
     "add_logger_name",
-    "add_extra",
+    "ExtraAdder",
     "BoundLogger",
     "filter_by_level",
     "get_logger",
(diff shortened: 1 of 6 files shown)


hynek:
Since you refer to 209 and also note yourself in #378 that adding all of callsite info would be overwhelming, I'd like to refer to [my comment](https://github.com/hynek/structlog/issues/209#issuecomment-966053578).

I think adding _all_ data ist just too much. We just have to decide if we want to take an allowlist or an disallowlist approach.

Since you've taken an allowlist approach in #380, I would say the same would be good here too? I usually like it better since it gives one more control.

We can't use enums of course, because we don't know what keys that user has added.

aucampia:
> Since you've taken an allowlist approach in #380, I would say the same would be good here too? I usually like it better since it gives one more control.
> 
> We can't use enums of course, because we don't know what keys that user has added.

Okay sounds like a good idea, I will change this to a processor class and add an allow list, and also provide an option for allowing copying of all values.

hynek:
Yes, that's even better!

hynek:
maybe just make it `allow=None` ==  all?

aucampia:
> maybe just make it `allow=None` == all?
> 
> i.e. `allow: Optional[Sequence]`

This will be simplest, I will go for this, will update PR tonight (CET)

aucampia:
Okay I converted it to a processor class, `ExtraAdder` and updated tests for this, still have to fix up documentation, will get to it tomorrow.

aucampia:
Documentation and changelog updated now also. I made the allow parameter a `Collection` instead of a `Sequence` so it can accept a set also.

src/structlog/stdlib.py
```diff
@@ -664,6 +667,66 @@ def add_logger_name(
...
+
+    .. versionadded:: 21.5.0
+    """
+
+    __slots__ = ["_copier"]
+
+    def __init__(self, allow: Optional[Collection[str]] = None) -> None:
+        self._copier: Callable[[EventDict, logging.LogRecord], None]
+        if allow is not None:
+            self._copier = functools.partial(self._copy_allowed, [*allow])
```
hynek, line 698: Given that `_copy_allowed` takes `Collection[str]` too, this conversion shouldn't be necessary, right?

```suggestion
            self._copier = functools.partial(self._copy_allowed, allow)
```

hynek (approved):
I've found a few nits, but they don't warrant another review round. This is great – thank you!

src/structlog/stdlib.py
hynek: (i'll leave it in, since the copy can save some surprises…)

hynek:
(I've merged it in https://github.com/hynek/structlog/commit/9208a2c7b39a85dd93b1625daeb48755d1d71573 because I couldn't push into your branch – thanks again!)

aucampia:
@hynek I will add tests for the two uncovered lines tonight (again CET), did not realize that test coverage target is 100% but should have added a test for case where there is no log record.

I will also add a comment for `[*allow]` instead of `allow` - this is mainly to prevent any surprises if the collection is mutated afterwards - probably it is a bit defensive to do that, but at the very least it is good to clarify why it is happening.

And please mention other issues if there are, will update them as I get time.

I will also make a PR to update PULL_REQUEST_TEMPLATE with:
 - consider giving you permission to their PR branches
 - ensure coverage is 100%

hynek:
I've fixed the coverage; feel free to open a hock PR for the allow thing tho!

hynek:
(i've also updated the template – sorry i've misread your comment on the go on a phone. i've only noticed you want to do it after looking up the PR id to refer to it)

PR #392, opened 2022-01-03
Narrow the return type of bind,unbind,try_unbind,new in subclass

### LINKED ISSUE #390: Narrow return types of `BindableLogger.{bind,unbind,try_unbind,new}` in `FilteringBoundLogger`
I usually use structlog like this:

```python
import structlog
from structlog.types import FilteringBoundLogger

logger: FilteringBoundLogger = structlog.get_logger(__name__)

log = cast(FilteringBoundLogger, logger.bind(something="value")
log.info("event description")
```

It would be nice to not have to cast the result of bind, one option for this would be to just override `{bind,unbind,try_unbind,new}` in `FilteringBoundLogger` to return `FilteringBoundLogger`.

I would be happy to make a PR for this, but this is somewhat of a commitment that I think would be best to discuss before a change is made, so please let me know if you are open to this. I started some work on it [here](https://github.com/hynek/structlog/compare/main...aucampia:iwana-20220103T0035-narrow_return_type), to show the concept.

aucampia:
# Summary

This narrows the return type of `bind`,`unbind`,`try_unbind` and `new`
in `FilteringBoundLogger` to return `FilteringBoundLogger` instead of
`BindableLogger`. This makes it easier to use structlog in a typed
context like this:

```python
def typecheck_filtering_return() -> None:
    fblogger: FilteringBoundLogger = structlog.get_logger(__name__)
    fblog = fblogger.bind(key1="value1", key2="value2", key3="value3")
    fblog.info("values bound")
    fblog = fblog.unbind("key1")
    fblog.debug("value unbound")
    fblog = fblog.try_unbind("bad_key")
    fblog.warn("no value unbound because key not defined")
    fblog = fblog.new(new="value")
    fblog.info("this is a whole new logger")
```

Without this change the above example will result in a type
error when using `mypy --strict` as the type of `fblog` will be `BindableLogger`
which does not have an `info` method:

```console
typing_examples.py:288: error: "BindableLogger" has no attribute "info"
typing_examples.py:290: error: "BindableLogger" has no attribute "debug"
typing_examples.py:292: error: "BindableLogger" has no attribute "warn"
typing_examples.py:294: error: "BindableLogger" has no attribute "info"
Found 4 errors in 1 file (checked 19 source files)
```

Fixes #390

# Pull Request Check List

- [x] Changes (and possible deprecations) are documented in the [**changelog**](https://github.com/hynek/structlog/blob/main/CHANGELOG.rst).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,7 @@ Whenever there is a need to break compatibility, it is announced here in the cha
 
 .. changelog
 
-XX.Y.Z (UNRELEASED)
+22.1.0 (UNRELEASED)
 -------------------
 
 Backward-incompatible changes:
@@ -36,7 +36,9 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-*none*
+- Overloaded the ``bind``, ``unbind``, ``try_unbind`` and ``new`` methods in the ``FilteringBoundLogger`` `Protocol <https://docs.python.org/3/library/typing.html#typing.Protocol>_`.
+  This makes it easier to use objects of type ``FilteringBoundLogger`` in a typed context.
+  `#392 <https://github.com/hynek/structlog/pull/392>`_
 
 
 ----
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -36,7 +36,7 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-- Overloaded the ``bind``, ``unbind``, ``try_unbind`` and ``new`` methods in the ``FilteringBoundLogger`` `Protocol <https://docs.python.org/3/library/typing.html#typing.Protocol>_`.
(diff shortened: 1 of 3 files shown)


aucampia:
More documentation may be appropriate or needed, but I'm not entirely sure where.

aucampia:
Also as said in the issue, this commitment to narrower return types in `FilteringBoundLogger` may not be something that you are comfortable with, in which case I'm fine if this PR is rejected. Would be nice to have but if there are use cases where it may make sense for a `FilteringBoundLogger` to not return `FilteringBoundLogger` from `{bind,unbind,try_unbind,new}` then it is better to not break them.

hynek (approved):
I honest to god can't come up with a use-case where that might be a problem and am willing to try it. :)

If someone comes up with it later, we can always revert it.

Thanks again1

GuillaumeDesforges:
Hi! Thanks @aucampia that's awesome!

Is a release planned to ship this feature to PyPI?

PR #399, opened 2022-02-09
PrintLogger: switch to print

Tinche:
Switch to print

### DIFF
--- a/src/structlog/_loggers.py
+++ b/src/structlog/_loggers.py
@@ -12,6 +12,7 @@ import sys
 import threading
 
 from pickle import PicklingError
+from sys import stderr, stdout
 from typing import IO, Any, BinaryIO, Dict, Optional, TextIO
 
 from structlog._utils import until_not_interrupted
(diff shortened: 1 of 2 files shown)


src/structlog/_loggers.py
```diff
@@ -12,6 +12,7 @@
 import threading
 
 from pickle import PicklingError
+from sys import stderr, stdout
```
hynek, line 15: Why this additionally to sys above?

Tinche: If you do `sys.stdout`, `print(file=None if self._file is sys.stdout else self._file` won't work properly when `sys.stdout` gets patched (by rich, for example).

Tinche: Because `self._file` will be unpatched, and `sys.stdout` will be `<rich.file_proxy.FileProxy object at 0x12407aac0>`. And also it's an additional attribute access, who wants that right?

Tinche:
Ok, ready for review.

src/structlog/_loggers.py
hynek: Could you please add a test that proves this property then?

Tinche: Yessir.

hynek:
OK thanks! The next structlog release is blocked now until you supply us with a `LowLevelWriter` as discussed in private. ;)

PR #401, opened 2022-02-22
fix: Use a noop with the same signature as the real one for make_filtering_bound_logger

GriceTurrble:
# Summary

In my team's app, we're using `make_filtering_bound_logger` in our configuration, which is working fine except in one corner case. Another developer introduced a bug that was missed in review and testing, and it only appears at runtime when the log level is DEBUG (below INFO, our standard level):

```py
log.debug(x=y)
```

Obviously this is incorrect: there needs to be an `event` string in this call. However, the original `_nop` method that gets returned has no checks for this missing argument, so their (admittedly poor) testing didn't turn up a problem at this line. Later when running at DEBUG level to research a different issue, the entire app failed to start up because of this call alone.

Root cause for us was incorrect code from our teammate, but I figured it could use some help from structlog, which seems to be silently passing what should be an error.

This change operates identically, returning `None` when the log level is too high. Now, though, it returns a method with the same signature every time, so a logging call written incorrectly will always raise an exception for missing arguments, regardless of the configured log level.

# Pull Request Check List

- [ ] **New APIs** are added to [`typing_examples.py`](https://github.com/hynek/structlog/blob/main/tests/typing_examples.py).
- [x] Changes (and possible deprecations) are documented in the [**changelog**](https://github.com/hynek/structlog/blob/main/CHANGELOG.rst).

### DIFF
--- a/src/structlog/_log_levels.py
+++ b/src/structlog/_log_levels.py
@@ -67,7 +67,7 @@ def add_log_level(
     return event_dict
 
 
-def _nop(*args: Any, **kw: Any) -> Any:
+def _nop(self: Any, event: str, **kw: Any) -> Any:
     return None
 
 
--- a/tests/test_log_levels.py
+++ b/tests/test_log_levels.py
@@ -34,11 +34,33 @@ class TestFilteringLogger:
 
     def test_one_below(self, bl, cl):
         """
-        if log level is exactly the min_level, log.
+        if log level is below the min_level, don't log.
         """
         bl.debug("nope")
 
         assert [] == cl.calls
+
+    def test_filter_bound_below_missing_event_string(self, bl, cl):
+        """
+        Missing event arg causes exception below min_level.
+        """
+        with pytest.raises(TypeError) as exc_info:
+            bl.debug(missing="event string!")
+        assert exc_info.type is TypeError
+
+        message = "missing 1 required positional argument: 'event'"
+        assert message in exc_info.value.args[0]
+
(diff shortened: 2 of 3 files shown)


tests/test_log_levels.py
```diff
@@ -34,12 +34,32 @@ def test_exact_level(self, bl, cl):
 
     def test_one_below(self, bl, cl):
         """
-        if log level is exactly the min_level, log.
+        if log level is below the min_level, don't log.
```
GriceTurrble, line 37: Noticed this docstring was incorrect, took the liberty to update :)

src/structlog/_log_levels.py
```diff
@@ -67,10 +67,6 @@ def add_log_level(
     return event_dict
 
 
-def _nop(*args: Any, **kw: Any) -> Any:
-    return None
```
GriceTurrble, line 71: Found no other uses of this method other than the one being changed below, so removed it.

hynek: I’m on vacation and phone-only, but is there a reason to pull the definition as a closure? That causes both runtime and memory overhead to create the same method over and over again?

GriceTurrble: Hadn't considered a memory impact, I figured the existing logic already made different methods if the min_level was low enough. But, fair point, I'll switch it back so the only change is the signature of `_nop`.

GriceTurrble:
Just came back around to this and squashed all the commits down to the most pertinent changes. :)

hynek (approved):
Thanks!

PR #404, opened 2022-03-21
Handle monkey-patched stdout

bbeattie-phxlabs:
# Summary

In some Python environments (e.g. Maya 2022's mayapy.exe), sys.stdout has been monkey-patched to something other than the expected stdout. As such, it doesn't actually have an `isatty` method and raises an AttributeError in response.


hynek (approved):
Man that expression is getting more and more fun!

Thanks!

PR #407, opened 2022-04-03
Add traceback parser for structured tracebacks

sscherfke:
# Summary

Parse tracebacks into an object structure than can easily be converted, e.g., to JSON. The code is based on *rich* (I talked to @willmcgugan and he is okay with that) but simplified and stripped of every "pretty printing" functionality.  It’s main goal is to produce structured tracebacks that can be parsed by log aggregators.

This is still work-in-progress. TODO:
- Documentation
- How to integrate it with structlog? I currently use an orjson processor that also calls the traceback parser (similar to what the ConsoleRenderer does with rich):

```python
import sys

import orjson
import structlog
from structlog.processors import _json_fallback_handler as orjson_fallback_handler
from structlog.types import EventDict, WrappedLogger


def render_orjson(_logger: WrappedLogger, _name: str, event_dict: EventDict) -> str:
    """
    Modified version of :class:`structlog.processors.JSONRenderer` that works around
    the issues in https://github.com/hynek/structlog/issues/360
    """
    exc_info = event_dict.pop("exc_info", None)
    if exc_info:
        event_dict["exception"] = get_traceback_dicts(exc_info)

    return orjson.dumps(event_dict, default=orjson_fallback_handler).decode()


def fail():
    1 / 0


def main():
    processors = [
        structlog.processors.add_log_level,
        structlog.processors.TimeStamper(),
    ]
    if sys.stdout.isatty():
        processors.append(structlog.dev.ConsoleRenderer())
    else:
        processors.append(render_orjson)

    structlog.configure(
        processors=processors,
    )
    log = structlog.get_logger()
    log.info("ohai")


    try:
        fail()
    except Exception as e:
        log.exception("onoes", exc_info=e)


if __name__ == "__main__":
    main()
```
Output:
```json
# python tb.py 2>&1 | jq
{
  "event": "ohai",
  "level": "info",
  "timestamp": 1648975014.737222
}
{
  "event": "onoes",
  "level": "error",
  "timestamp": 1648975014.7372942,
  "exception": [
    {
      "exc_type": "ZeroDivisionError",
      "exc_value": "division by zero",
      "syntax_error": null,
      "is_cause": false,
      "frames": [
        {
          "filename": "/Users/stefan/Projects/typed-settings/tb.py",
          "lineno": 280,
          "name": "main",
          "line": "",
          "locals": {
            "processors": "'[<function add_log_level at 0x109a1dfc0>, <structlog.processors.TimeStamper obje'+60",
            "log": "'<BoundLoggerLazyProxy(logger=None, wrapper_class=None, processors=None, context_'+55",
            "e": "ZeroDivisionError('division by zero')"
          }
        },
        {
          "filename": "/Users/stefan/Projects/typed-settings/tb.py",
          "lineno": 259,
          "name": "fail",
          "line": "",
          "locals": {}
        }
      ]
    }
  ]
}
```

# Pull Request Check List


hynek:
Is there anything concrete you need from _me_?

sscherfke:
I think the main question is how to integrate it into structlog.  The rich-traceback-handling is "hardcoded" into the ConsoleLogger, so I was thinking about somehting like
```python
def render_orjson(_logger: WrappedLogger, _name: str, event_dict: EventDict) -> str:
    """
    Modified version of :class:`structlog.processors.JSONRenderer` that works around
    the issues in https://github.com/hynek/structlog/issues/360
    """
    exc_info = event_dict.pop("exc_info", None)
    if exc_info:
        event_dict["exception"] = get_traceback_dicts(exc_info)

    return orjson.dumps(event_dict, default=orjson_fallback_handler).decode()
```
But that would then be coupled to much with orjson.

The exc-info-processor would not work since it "breaks" pretty-rich-tracebacks when you it.

But I did not yet have enough time to look for a good solution.

hynek:
Given how unforthcoming orjson has been, I'd rather not have specific APIs for it in structlog.

If `event_dict["exception"]` is a dict, why does it need special handling? Isn't this just a case of https://www.structlog.org/en/stable/api.html#structlog.processors.format_exc_info ? The console code is hardcoded because we use rich for formatting…

sscherfke:
@hynek I prototyped a new approach:

`format_exc_info()` is now customizable with a formatter function. I created a new class than can be configured with a custom formatter. `format_exc` is an alias to an instance of that class with the current formatter.  That should assure backwards compatibility.

The JSON format can now be used like this:

```
from structlog import processors, tracebacks

format_json = tracebacks.JsonFormatter(...)
procs = [
    ...,
    processors.ExceptionFormatter(format_json),
    ...,
]
```

What do you think?  Shall I continue this way?

src/structlog/processors.py
```diff
@@ -289,7 +294,7 @@ class JSONRenderer:
     """
     Render the ``event_dict`` using ``serializer(event_dict, **json_kw)``.
 
-    :param json_kw: Are passed unmodified to *serializer*.  If *default*
+    :param dumps_kw: Are passed unmodified to *serializer*.  If *default*
```
hynek, line 297: I took the freedom to fix this in main. :)

src/structlog/tracebacks.py
```diff
@@ -0,0 +1,222 @@
...
+from typing import Any, Optional, Type, Union
+
+from .types import ExcInfo
+
+
+SHOW_LOCALS = True
+LOCALS_MAX_STRING = 80
+MAX_FRAMES = 50
+
+ExcInfo = tuple[Type[BaseException], BaseException, Optional[TracebackType]]
```
hynek, line 22: this notation only works in annotations unfortunately. you'll have to import `Tuple`

src/structlog/processors.py
```diff
@@ -360,13 +367,40 @@ def format_exc_info(
...
+            )
+
+        return event_dict
+
+
+format_exc_info = ExceptionFormatter()
+"""
+Replace an ``exc_info`` field with an ``exception`` string field:
+
+If *event_dict* contains the key ``exc_info``, there are two possible
```
hynek, line 393: off by 1? ;)

src/structlog/processors.py
```diff
@@ -360,13 +367,40 @@ def format_exc_info(
...
+behaviors:
+
+- If the value is a tuple, render it into the key ``exception``.
+- If the value is an Exception render it into the key ``exception``.
+- If the value true but no tuple, obtain exc_info ourselves and render
```
hynek, line 398:
```suggestion
- If the value is true but no tuple, obtain exc_info ourselves and render
```

hynek:
I think that indirection makes sense, yes. We'll have some work to do around `ExceptionGroup` too. :-/

src/structlog/tracebacks.py
sscherfke: Yes, you are right. But it was just a prototype to demonstrate my ideas, though ;-)

sscherfke:
> We'll have some work to do around `ExceptionGroup` too. :-/

What exactly is the issue with `ExceptionGroup`? I haven't gotten into that yet.

hynek:
@Tinche complained about something…I think they got rendered line by line or something?

src/structlog/processors.py
sscherfke: This is just a copy of *your* comment :-P
But I will fix it for you ;-)

sscherfke:
Woop woop. The tests are finally passing.  Need to write some docs now.  When this is done, I’ll rebase on main.

sscherfke:
Maybe I’m done now, but probaply I missed something. 🤷

sscherfke:
I will add a `json_exceptions = ExceptionFormatter(tracebacks.JSONExceptionFormatter())` (similar to `format_exc_info()`) shortcut which makes using it with the jsonrenderer more convenient.

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ So please make sure to **always** properly configure your applications.
   [#404](https://github.com/hynek/structlog/pull/404)
 - `structlog.make_filtering_bound_logger()`-returned loggers now also have a `log()` method to match the `structlog.stdlib.BoundLogger` signature closer.
   [#413](https://github.com/hynek/structlog/pull/413)
+- Added structured logging of tracebacks via the `structlog.tracebacks` module,
+  and most notably the `structlog.tracebacks.JSONFormatter` which can be used with the new `structlog.processors.ExceptionFormatter`. [#407](https://github.com/hynek/structlog/pull/407)
```
hynek, line 51:
```suggestion
  and most notably the `structlog.tracebacks.JSONFormatter` which can be used with the new `structlog.processors.ExceptionFormatter`.
  [#407](https://github.com/hynek/structlog/pull/407)
```

src/structlog/tracebacks.py
```diff
@@ -0,0 +1,261 @@
...
+            continue
+
+        # No cover, code is reached but coverage doesn't recognize it.
+        break  # pragma: no cover
+
+    trace = Trace(stacks=stacks)
+    return trace
+
+
+class JSONFormatter:
```
hynek, line 200: So it just occurred to me that this formatter (and everything related to it) shouldn't carry the JSON name, because it's misleading.

While in 99.9% of cases the structured exceptions are gonna be serialized into JSON, it doesn't format it to JSON!

I realize it's gonna be somewhat less-obvious to document, but OTOH we'd be underselling it, if we'd call it this way.

Maybe `StructuredFormatter`? I'm open to better names, but JSON is shouldn't be it.

CHANGELOG.md
```diff
@@ -58,6 +60,10 @@ So please make sure to **always** properly configure your applications.
   [#399](https://github.com/hynek/structlog/pull/399)
 - `structlog.testing.capture_logs` now works for already initialized bound loggers.
   [#408](https://github.com/hynek/structlog/pull/412)
+- `structlog.processors.format_exc_info()` is no longer a function but
+  an instance of `structlog.processors.ExceptionFormatter`.
+  Its behavior has not been changed.
```
hynek, line 65: Active voice ftw

```suggestion
- `structlog.processors.format_exc_info()` is no longer a function, but an instance of `structlog.processors.ExceptionFormatter`.
  Its behavior has not changed.
```

hynek (changes requested):
Sorry, one **BIG** bikeshed that only now occurred to me.

Tinche:
So I asked for ExceptionGroup support some weeks ago but I withdraw my request. The ExceptionGroups render their traceback using some ASCII art which looks funky embedded in JSON but looks completely OK when inspected in ElasticSearch.

I wouldn't parse my errors into structured data since it's unnecessary work for the service and the traceback is searchable just fine as text in ES. Just my context, other people might find it very useful ;)

src/structlog/tracebacks.py
sscherfke: Screw you! I was thinking the same and either had no better name.  So I hoped I could get away with `JSONFormatter` 🤪

I’ll try to hit Ballmer's Peek tonight and see if I can come up with a better name.  😬

sscherfke: Maybe just `DictFormatter` since it returns a list of trace dicts ...

hynek: I was gonna suggest DictFormatter but was detracted by the fact that it returns a list...but I guess it works anyways and is nicer to type/say.

sscherfke: What about `json_tracebacks`?  Whould we rename it to `dict_tracebacks` accordingly? But OTOH it looks quite good as it is in its usage context:
```python
         ...     processors=[
         ...         structlog.processors.json_tracebacks,
         ...         structlog.processors.JSONRenderer(),
         ...     ],
```

hynek: It looks good, but it's still wrong. :)

sscherfke: done

docs/api.rst
```diff
@@ -170,6 +170,27 @@ API Reference
...
+         ...         structlog.processors.dict_tracebacks,
+         ...         structlog.processors.JSONRenderer(),
+         ...     ],
+         ... )
+         >>> log = structlog.get_logger()
+         >>> var = "spam"
+         >>> try:
+         ...     1 / 0
+         ... except ZeroDivisionError as e:
+         ...     log.exception("Cannot compute!", exc_info=e)
```
hynek, line 191: I _think_ this should work:

```suggestion
         ...     log.exception("Cannot compute!")
```

src/structlog/processors.py
```diff
@@ -41,7 +41,13 @@
 )
 from ._log_levels import _NAME_TO_LEVEL, add_log_level
 from ._utils import get_processname
-from .types import EventDict, ExcInfo, WrappedLogger
+from .tracebacks import DictFormatter as DictExceptionFormatter
```
hynek, line 44: Why the `as`?

src/structlog/processors.py
```diff
@@ -343,13 +350,20 @@ def _json_fallback_handler(obj: Any) -> Any:
...
-def format_exc_info(
-    logger: WrappedLogger, name: str, event_dict: EventDict
-) -> EventDict:
+class ExceptionFormatter:
     """
-    Replace an ``exc_info`` field by an ``exception`` string field:
+    Replace an ``exc_info`` field with an ``exception`` field which is rendered
+    by *exception_formatter*.
+
+    The contents of ``exception`` field depend on the return value of the
```
hynek, line 358:
```suggestion
    The contents of the ``exception`` field depends on the return value of the
```

src/structlog/processors.py
```diff
@@ -359,14 +373,69 @@ def format_exc_info(
...
+
+        return event_dict
+
+
+format_exc_info = ExceptionFormatter()
+"""
+Replace an ``exc_info`` field with an ``exception`` string field using
+Python's built-in traceback formatting.
+
+If *event_dict* contains the key ``exc_info``, there are tree possible
```
hynek, line 406: 😂🌲

```suggestion
If *event_dict* contains the key ``exc_info``, there are three possible
```

src/structlog/processors.py
```diff
@@ -359,14 +373,69 @@ def format_exc_info(
...
+behaviors:
+
+- If the value is a tuple, render it into the key ``exception``.
+- If the value is an Exception render it into the key ``exception``.
+- If the value is true but no tuple, obtain exc_info ourselves and render
+    that.
```
hynek, line 412: let's enumerate when we mention numbers:

```suggestion
1. If the value is a tuple, treat it like a tuple coming from `sys.exc_info`
   and render it into the key ``exception``.
2. If the value is an Exception, render it into the key ``exception``.
3. If the value is true but no tuple, obtain `sys.exc_info` ourselves and render
    that.
```

src/structlog/processors.py
```diff
@@ -359,14 +373,69 @@ def format_exc_info(
...
+
+If there is no ``exc_info`` key, the *event_dict* is not touched.
+This behavior is analogue to the one of the stdlib's logging.
```
hynek, line 435: Let's dedup this

```suggestion
The treatment of the ``exc_info`` key is identical to `format_exc_info`.
```

src/structlog/types.py
```diff
@@ -97,6 +97,30 @@
 """
 
 
+@runtime_checkable
+class ProcessorExceptionFormatter(Protocol):
+    """
+    **Protocol:** A callable that formats an `ExcInfo` into another
+    datastructure.
+
+    The result should be something, that your renderer can work with, e.g., a
```
hynek, line 106: my favorite fellow-German-speaker error that I can't stop making

```suggestion
    The result should be something that your renderer can work with, e.g., a
```

src/structlog/types.py
```diff
@@ -97,6 +97,30 @@
...
+    ``str`` or a JSON serializable ``dict``.
```
hynek, line 107: compound adjective

```suggestion
    ``str`` or a JSON-serializable ``dict``.
```

src/structlog/types.py
```diff
@@ -97,6 +97,30 @@
...
+
+    Used by `structlog.processors.format_exc_info()` and
+    `structlog.processors.ExceptionPrettyPrinter`.
+
+    :param exc_info: Is the exception tuple to format
+
+    :returns: Anything that can be rendered by the last processor in your
+        change, e.g. a string or a JSON structure.
```
hynek, line 115:
```suggestion
        chain, e.g. a string or a JSON-serializable structure.
```

tests/test_processors.py
```diff
@@ -491,25 +500,21 @@ def test_gets_exc_info_on_bool(self):
...
         except ValueError:
-            d = format_exc_info(None, None, {"exc_info": True})
+            d = ExceptionFormatter()(None, None, {"exc_info": True})
 
         assert "exc_info" not in d
         assert 'raise ValueError("test")' in d["exception"]
         assert "ValueError: test" in d["exception"]
 
-    def test_exception_on_py3(self, monkeypatch):
+    def test_exception_on_py3(self):
```
hynek, line 509: While you're touching it :)

```suggestion
    def test_exception(self):
```

hynek (approved):
bunch of nitpicks but i think we're mostly ready to merge!

relying a bit on the fact that you're already using it successfully in prod 😇

src/structlog/processors.py
sscherfke: To make it clear for users of that package, that `DictFormatter` is not a log-formatter, but an exception formatter.

`DictFormatter` is a good name if you take its module name into account, but with `from` imports, this information is lost.

But if you want the `as ...` to be removed, I will remove it. :)

src/structlog/types.py
sscherfke: 🙈

src/structlog/processors.py
hynek: I would suggest to call the formatter DictExceptionFormatter. I know that structlog is a bit wild on not doing stutter imports (we've got like 2094 `BoundLogger`s), but in hindsight it was a mistake I think.

src/structlog/types.py
```diff
@@ -97,6 +97,30 @@
 """
 
 
+@runtime_checkable
+class ProcessorExceptionFormatter(Protocol):
```
hynek, line 101: ugh OK now I'm realizing we've gotten ourselves into a mess naming-wise :-/

We've got three things ending in "Formatter":

- ProcessorExceptionFormatter, a protocol
- DictExceptionFormatter, something that formats exc_info into dicts
- ExceptionFormatter, something that replaces exc_info by exceptions (processed potentially by DictExceptionFormatter)

We need to improve this. 😅

My gut feeling tells me that actually _none_ of them should be called a formatter.

- DictExceptionFormatter doesn't format anything, it returns a list of dicts.
- same argument applies to ProcessorExceptionFormatter, simply by virtue of the existence of DictExceptionFormatter

I suspect Renderer would be a better name here, since we use it for JSONRenderer too? I would appreciate your thoughts here tho.

A Formatter-suffix would be a good name for a class that formats to text, as the old processor did, but that doesn't seem to have an own class.

I would postpone the discussion about the naming of ExceptionFormatter until we've nailed the other two down

hynek (changes requested):
sorry, more thinking required 😅

I suspect your gut feeling was telling that you too before ;)

src/structlog/types.py
sscherfke: Just my first thoughts, need to sleep over them:

- `ExceptionFormatter` -> renders exceptions into something that can be printed by the last processor -> `ExceptionRenderer`
- `DictExceptionFromatter` -> transforms a traceback from a long string to dicts -> `ExceptionDictTransformer`
- `ProcessorExceptionFormatter` -> A protocol for exception transformers used by the exception renderer -> `ExceptionTransformer`

hynek: that's one looooong nap ;)

I think I'm OK with those names unless your super-nap unveiled some space brains. :)

sscherfke: No, several naps did not lead to better names. 

I will change the code to these names. Then we can get a feeling of how they feel. 👌

hynek (approved):
All right, I think we've got it!

Thanks!

PR #409, opened 2022-04-09
Deprecate structlog.threadlocal

### LINKED ISSUE #406: structlog.threadlocal should be deprecated
I didn't realize it originally, but [as confirmed by Yury](https://twitter.com/1st1/status/1483807086341406720), contextvars are a higher-level abstraction that includes threadlocal and therefore we should stop maintaining both.

Greenlets appear to be [supported too](https://greenlet.readthedocs.io/en/latest/contextvars.html), so we can effectively stop maintaining/testing that code.

---

I'm afraid of the shitstorm if we removed it, but I would deprecate it, add a fat banner in the docs steering people to contextvars, and simplify our testing matrix.

cc @MarkusH

hynek:
Fixes #406

cc @MarkusH @radix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ So please make sure to **always** properly configure your applications.
 
 - The entire `structlog.threadlocal` module is deprecated.
   Please use the primitives from `structlog.contextvars` instead.
+  [#409](https://github.com/hynek/structlog/pull/409)
 
 
 ### Added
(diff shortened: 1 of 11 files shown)


radix (approved):
LGTM. I guess it's time for me to switch :)

hynek:
Thanks Chris, and especially thank you for introducing these APIs! They were a game changer and I use them in all my projects now (via contextvars)!

ionelmc:
Sorry for silly question here: merge_threadlocal is redundant if I already use merge_contextvars right?

hynek:
yes, but they don't work on each other's data.

i.e. merge_contextvars only merges whatever has been bound using bind_contextvars and vice versa

PR #412, opened 2022-04-19
Apply capture_logs to initialized bound loggers

### LINKED ISSUE #408: `capture_logs` doesn't affect already initalized `BoundLogger`s
Hey, I found an issue with using `structlog.testing.capture_logs` that might be intentional, but tripped me up while testing.
I have a workaround / patch that I could submit as PR, but wanted to ask first to clarify whether the behavior is intentional.

When using `capture_logs` the processors of any existing, already initialized `BoundLogger`s aren't updated to use `LogCapture` and their logs are thus not captured.

I've written a bit of code to clarify what I mean:

```python
from contextlib import contextmanager
from typing import Iterator, List

import structlog
from structlog.dev import ConsoleRenderer
from structlog.processors import add_log_level
from structlog.testing import capture_logs
from structlog.types import EventDict

processors = [add_log_level, ConsoleRenderer()]
structlog.configure(processors=processors)

# processors list is stored by reference
structlog.get_config()['processors'] is processors

logger = structlog.get_logger('log1')

# initialize BoundLogger by using the logger
logger = logger.bind(foo='bar')
logger.info("baz")

# for now, the logger's processors are _identical_ to the configuration's
assert logger._processors is processors

with capture_logs() as events:
    # the logger's processors are still _identical_ to the original processors list
    assert logger._processors is processors
    assert len(processors) == 2
    # while the configuration's list has been exchanged
    assert structlog.get_config()['processors'] is not processors
    assert len(structlog.get_config()['processors']) == 1

    # thus logs won't be captured
    logger.info("test")
    assert len(events) == 0


@contextmanager
def capture_logs_by_ref() -> Iterator[List[EventDict]]:
    """Modify `capture_logs` to keep reference to `processors` list intact,
    in order to also modify bound loggers which get the list assigned by reference.

    """
    cap = structlog.testing.LogCapture()
    # Modify `_Configuration.default_processors` set via `configure` but always keep
    # the list instance intact to not break references held by bound loggers.
    processors = structlog.get_config()["processors"]
    old_processors = processors[:]  # copy original processors
    try:
        processors.clear()  # clear processors list
        processors.append(cap)  # append the LogCapture processor for testing
        structlog.configure(processors=processors)
        yield cap.entries
    finally:
        processors.clear()  # remove LogCapture
        processors.extend(old_processors)  # restore original processors
        structlog.configure(processors=processors)


# let's try again
structlog.configure(processors=processors)
# processors list is stored by reference
structlog.get_config()['processors'] is processors

logger = structlog.get_logger('log2')

# initialize BoundLogger by using the logger
logger = logger.bind(foo='bar')
logger.info("baz")

# the logger's processors are still _identical_ to the configuration's
assert logger._processors is processors

with capture_logs_by_ref() as events:
    # the logger's processors are still _identical_ to the original processors list
    assert logger._processors is processors
    assert structlog.get_config()['processors'] is processors
    # but it contains only the LogCapture processor
    assert len(processors) == 1

    logger.info("test")
    assert events[0]['event'] == "test"

# and after exiting the capture_logs context, the original processors are restored
assert len(processors) == 2
```

That basically already contains my approach for a fix: keeping the `_Configuration.processors` list intact and only modify its contents so that the `BoundLogger`s which get the list passed by reference are affected by the changes.

Let me know if you agree that this change would be benefitial, then I can prepare a PR with tests.

And of course: thanks for building structlog - it's amazing!

fabianbuechler:
# Summary

This changes the `structlog.testing.capture_logs` contextmanager to also work for already initialized bound loggers.

Fixes #408.

# Pull Request Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@ So please make sure to **always** properly configure your applications.
   [#401](https://github.com/hynek/structlog/pull/401)
 - `structlog.PrintLogger` -- that is used by default -- now uses `print()` for printing, making it a better citizen for interactive terminal applications.
   [#399](https://github.com/hynek/structlog/pull/399)
+- `structlog.testing.capture_logs` now works for already initialized bound loggers.
+  [#408](https://github.com/hynek/structlog/pull/412)
 
 
 ### Fixed
--- a/src/structlog/testing.py
+++ b/src/structlog/testing.py
@@ -59,12 +59,22 @@ def capture_logs() -> Generator[List[EventDict], None, None]:
     .. versionadded:: 20.1.0
     """
     cap = LogCapture()
-    old_processors = get_config()["processors"]
+    # Modify `_Configuration.default_processors` set via `configure` but always
+    # keep the list instance intact to not break references held by bound
+    # loggers.
+    processors = get_config()["processors"]
+    old_processors = processors.copy()
     try:
-        configure(processors=[cap])
+        # clear processors list and use LogCapture for testing
+        processors.clear()
+        processors.append(cap)
+        configure(processors=processors)
         yield cap.entries
     finally:
-        configure(processors=old_processors)
+        # remove LogCapture and restore original processors
+        processors.clear()
+        processors.extend(old_processors)
+        configure(processors=processors)
 
 
 class ReturnLogger:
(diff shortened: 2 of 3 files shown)


hynek (approved):
Great, thanks!

PR #413, opened 2022-04-21
add log() to BoundLoggerFiltering*

FlorianLudwig:
# Summary

add `log()` to `BoundLoggerFiltering*` to better match `BoundLogger` signature 

# Pull Request Check List

- [?] **New APIs** are added to [`typing_examples.py`](https://github.com/hynek/structlog/blob/main/tests/typing_examples.py) (not actually new, just fixes `stdlib.get_logger().log()`)
    - [x] I think no documentation update is needed for this addition

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,12 +41,13 @@ So please make sure to **always** properly configure your applications.
   [#403](https://github.com/hynek/structlog/pull/403)
   [#404](https://github.com/hynek/structlog/pull/404)
 
-
 ### Changed
 
 - `structlog.make_filtering_bound_logger()` now returns a method with the same signature for all log levels, whether they are active or not.
   This ensures that invalid calls to inactive log levels are caught immediately and don't explode once the log level changes.
   [#401](https://github.com/hynek/structlog/pull/401)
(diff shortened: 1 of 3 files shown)


hynek:
When you say BoundLogger, I presume you mean stdlib's one (yes, calling them all same was stupid)?

There's really no guarantee to mirror stdlib APIs in general, do you find that method practically useful and if so how? Because it completely obliterates the performance gains we got by pre-creating concrete classes.

FlorianLudwig:
> When you say BoundLogger, I presume you mean stdlib's one (yes, calling them all same was stupid)?

Indeed. I like using `structlog.stdlib.get_logger` for the type hints.

> There's really no guarantee to mirror stdlib APIs in general, do you find that method practically useful and if so how?

I use it to pipe stdlib logs into structlog.

> Because it completely obliterates the performance gains we got by pre-creating concrete classes.

It should only effect performance for calls that use `.log` but the `.info` etc should be still the same performance, right?

hynek:
OK thanks! I'll fix a few tiny quibbles myself, really weird to document this. 🤔

tjb9dc:
@hynek would you mind cutting a release that has this PR? Thanks!

hynek:
I plan on pushing 22.1 once #407 is merged.

PR #414, opened 2022-04-22
Fix configuration example for stdlib

zifot:
# Summary

Fixes configuration sample in "Standard Library Logging" documentation page. At least as of 21.5.0 copy-pasting it required changes for it to work,

# Pull Request Check List


hynek:
Ah right that explains a few questions I've got!

Would you mind to change it to a set, since we use set in our defaults and it would be more consistent that way.

(just replace () for {})

Thank you!

hynek:
Thanks!

PR #449, opened 2022-09-30
Fix typo on `why.rst`

Kludex:
Hi 👋 

I'm actually not sure if this is a typo. Is it? If it's not, sorry. 👀


hynek (approved):
Yes! Great; thanks!

PR #451, opened 2022-09-30
Update getting-started.rst

Kludex:
It took me around 3 minutes to understand the sentence, maybe because I'm tired... But perhaps my suggestion makes it a bit clear? :thinking:


hynek:
Thanks, I've need another 2 minutes to understand even with your improvements, so I've dumbed it down a bit even further.

Issue #488, opened 2023-02-09
`aexception` doesn't pass the exception info


codepainters:
This simple test code
```python
import asyncio
import structlog
log = structlog.get_logger()


async def kaboom():
    try:
        print(1 / 0)
    except Exception:
        await log.aexception('kaboomed')


if __name__ == '__main__':
    asyncio.new_event_loop().run_until_complete(kaboom())
```
...fails with...
```
  // top of stack removed for brevity
  File "/home/czajnik/.cache/pypoetry/virtualenvs/poor-xCdm2f7w-py3.11/lib/python3.11/site-packages/structlog/dev.py", line 424, in __call__
    self._exception_formatter(sio, exc_info)
  File "/home/czajnik/.cache/pypoetry/virtualenvs/poor-xCdm2f7w-py3.11/lib/python3.11/site-packages/structlog/dev.py", line 187, in rich_traceback
    Traceback.from_exception(*exc_info, show_locals=True)
  File "/home/czajnik/.cache/pypoetry/virtualenvs/poor-xCdm2f7w-py3.11/lib/python3.11/site-packages/rich/traceback.py", line 335, in from_exception
    rich_traceback = cls.extract(
                     ^^^^^^^^^^^^
  File "/home/czajnik/.cache/pypoetry/virtualenvs/poor-xCdm2f7w-py3.11/lib/python3.11/site-packages/rich/traceback.py", line 405, in extract
    exc_type=safe_str(exc_type.__name__),
                      ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute '__name__'. Did you mean: '__ne__'?

Process finished with exit code 1
```

I believe the problem is here: https://github.com/hynek/structlog/blob/4e7bf267ab7a86b8b004cc76bbe60d3b41655346/src/structlog/_log_levels.py#L88-L95

Code spawned in the executor can no longer retrieve original exception information.

As a workaround I do the following:
```python
        await log.aexception('kaboomed', exc_info=sys.exc_info())
```

But why can't it be a default value for `exc_info`?

SerGeRybakov:
@codepainters @hynek, hi! 
I use 22.3.0 and have the same problem with no exc_info in aexception.
Is that merge #489 already released in any version or not? If still not, when it shall be released?
Thanks!

hynek:
23.1.0 is on PyPI

PR #489, opened 2023-02-09
Fix missing exception info in aexception()

codepainters:
# Summary

This is how I've solved #488 for my needs.

I've not updated the test cases. I can do it, if you're happy with this change.

### DIFF
--- a/src/structlog/_log_levels.py
+++ b/src/structlog/_log_levels.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import asyncio
 import contextvars
 import logging
+import sys
 
 from typing import Any, Callable
 
@@ -86,7 +87,10 @@ def exception(self: FilteringBoundLogger, event: str, **kw: Any) -> Any:
 
 
 async def aexception(self: FilteringBoundLogger, event: str, **kw: Any) -> Any:
-    kw.setdefault("exc_info", True)
+    # Exception info has to be extracted this early, because it is no longer
+    # available once control is passed to the executor.
+    if kw.get("exc_info", True) is True:
+        kw["exc_info"] = sys.exc_info()
 
     ctx = contextvars.copy_context()
     return await asyncio.get_running_loop().run_in_executor(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,9 @@ You can find out backwards-compatibility policy [here](https://github.com/hynek/
   This prevents crashes if the actual Exception is passed for the *exc_info* argument instead of a tuple or `True`.
   [#482](https://github.com/hynek/structlog/issues/482)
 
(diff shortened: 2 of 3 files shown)


src/structlog/_log_levels.py
```diff
@@ -86,7 +87,8 @@ def exception(self: FilteringBoundLogger, event: str, **kw: Any) -> Any:
 
 
 async def aexception(self: FilteringBoundLogger, event: str, **kw: Any) -> Any:
-    kw.setdefault("exc_info", True)
+    if "exc_info" not in kw:
```
hynek, line 90: Given your explanation on #488, I think this should also handle the case of someone passing `True` here. It works everywhere else, so it's kinda unobvious that it doesn't here.

It would be also nice if you could add a comment explaining what's happening, so I don't "simplify" it in the future. ;)

hynek:
yeah the change makes sense plus what I commented inline. Please add a test & changeling entry. :)

src/structlog/_log_levels.py
codepainters: Actually I'm slightly puzzled about the expected behavior of this call.

Synchronous `exception()` either passes `exc_info` from `kwargs` as is, or sets it to `True` (which is replaced with actual exception info later on).  

For async variant, my assumption was to put use `sys.exc_info()` for default case, and pass everything else as is. Indeed, it makes explicit `aexception(...., exc_inf=True)` not working as expected.

However - should I replace literal `True` with `exc_info`? Replacing any true-ish value doesn't seem right, either.

hynek: yes, that's what I meant: replace `True` (and only `True`) by `sys.exc_info()` in addition to your logic.

I.e. something like `if kw.get("exc_info", True) is True:`:

```pycon
>>> {}.get("exc_info", True) is True
True
>>> {"exc_info": True}.get("exc_info", True) is True
True
>>> {"exc_info": (1,2,3)}.get("exc_info", True) is True
False
```

In vanilla `exception()`, passing True the way to tell structlog to figure it out, but as you've pointed out yourself: that's impossible once we're in a new task so we have to handle it here.

codepainters: Ok, makes sense. Update pushed.

hynek (approved):
Cool, thanks!

codepainters:
Let me ask - how come you had to pacify `flake8` ? Is the CI config different from pre-commit one ?

hynek:
> Let me ask - how come you had to pacify `flake8` ? Is the CI config different from pre-commit one ?

Not sure, it complained about comparing type() to a type instead of `isinstance`.

hynek:
(the mypy error is unrelated; I'll quickly fix it in main and rebase)

hynek:
Big success, thanks!

Issue #497, opened 2023-03-16
Mermaid Diagram rendering broken


schewara:
Hi

The Website only shows a blob of Text instead of the renderd Mermaid Diagrams.

hynek:
it should be working again

hynek:
@PeterJCLaw it should work again

PR #519, opened 2023-06-09
Fix Mermaid diagrams for Mermaid 10

hynek:
looks it's impossible to change the theme with Mermaid 10 without introducing the JavaScript build hell @mgaitan? :(

ref #497


hynek:
Figured it out!

Issue #530, opened 2023-07-15
FilteringBoundLogger exception() method has different API 


alexcouper:
FilteringBoundLogger `exception` call fails with positional arguments.

# Example
```python
import structlog
from structlog import get_logger
from structlog.stdlib import LoggerFactory

structlog.configure(logger_factory=LoggerFactory())
log = get_logger("test")
log.error("%s %s %s", "bob", "alice", "fred")
log.exception("%s %s %s", "bob", "alice", "fred")
```

Expected output:
```
2023-07-15 11:30:42 [error    ] bob alice fred
2023-07-15 11:30:42 [error    ] bob alice fred
```

Actual output:
```
2023-07-15 11:30:42 [error    ] bob alice fred
Traceback (most recent call last):
  File "/Users/alex/Work/syndis/aftra/debug/2023-07-13-logging.py", line 30, in <module>
    log.exception("%s %s %s", "bob", "alice", "fred")
TypeError: exception() takes 2 positional arguments but 5 were given
```

# Some historical digging
https://github.com/hynek/structlog/commit/b71aabb4f440dbd50b389995746aad3bcefa4160#diff-b87fe76f9eed7fd535af19eade4ac979499ddaaaba4611af36cc59f4b7f247fc introduced the `exception` method without the `*args` part of the signature.

That implementation was also later used  for the `exception` method used by `FilteringBoundLogger` [here](https://github.com/hynek/structlog/commit/bad0116427286c59e7d85a0ef16e0db663f8ea1d#diff-d3c48fb25e23fb8a3e2bb689675f8244e6225540d55bb1df2e3a6a81d4ee3fafR68)

A later fix was made to exception to "add pos arg support to exception" [here](https://github.com/hynek/structlog/commit/ca8c714965d37c3cad4e8de1e39e5bb9e4128fb0)

But no corresponding fix was made to the functions found in `_log_levels.py`.

# Question

Any reason I shouldn't provide a fix, that adjusts the `_log_levels.py` arguments to include `*args` and pass them through to the error call?

alexcouper:
This is the minimum set of configuration needed to make %s style formatting work with the `.exception` method (afaict):

```python
structlog.configure(
    processors=[
        structlog.stdlib.PositionalArgumentsFormatter(),
        structlog.processors.format_exc_info,
        structlog.stdlib.render_to_log_kwargs,
    ],
    logger_factory=structlog.stdlib.LoggerFactory(),
[...]

PR #531, opened 2023-07-15
Allow positional args in `exception`

alexcouper:
Possible solution for https://github.com/hynek/structlog/issues/530


hynek:
Thanks!

PR #535, opened 2023-07-27
Update JSONRenderer docs

jerr0328:
# Summary

I noticed *json_kw* was referred to in docs but the the actual parameter is called *dumps_kw*. While fixing that I also wanted to add a little bit of info I had to figure out on my own, which was that if you want `orjson` to handle serialization of dataclasses, datetimes, etc. natively, you need to pass a `default` (I used the `singledispatch` to use `str` by default and handle Pydantic objects and sets to work better with our logging systems)

# Pull Request Check List

Do I need to add to the changelog? 🤔 

- [x] **New APIs** are added to [`api.py`](https://github.com/hynek/structlog/blob/main/tests/typing/api.py).


hynek (approved):
Thanks! Who knows how long that was in there. :)

PR #539, opened 2023-08-05
Make processor chain more robust (representable state/typing)

alexpovel:
# Summary

This is a very rough first sketch. The change would touch deeper aspects of the code base and require some work. I didn't want to pull all that off only for the idea to then be rejected. Hence, the checklist below is empty: everything is broken, sorry! I still wanted to submit actual code and not just an issue as that's easiest to reason about and explain with. Here goes.

The [type system is wonderful](https://dusted.codes/the-type-system-is-a-programmers-best-friend) and `mypy` makes it better constantly for the Python ecosystem. I've become a huge fan of [compile-time guarantees/correctness](https://cliffle.com/blog/rust-typestate/) and pushing as much logic as possible into the type system, which is now possible in modern Python (always has been, but *checking* it hasn't). The "compiler" (mypy) is excellent at checking invariants etc. It can make [invalid states unrepresentable](https://geeklaunch.io/blog/make-invalid-states-unrepresentable/#legal-vs-valid-states). Bla bla bla... you're aware of all that. This package is typed already, and strictly so, which is excellent.

However, when using structlog, I noticed the processor chain can be misused without the type system being able to warn users about it:

```python
import re

import structlog

from structlog.processors import JSONRenderer, TimeStamper
from structlog.typing import EventDict, WrappedLogger


def scrub(
    logger: WrappedLogger,
    name: str,
    # ↓ annotation contradicts docs, and signature of `structlog.typing.Processor`, but
    # works as intended
    event_dict: EventDict | str,
) -> str:
    """Removes sensitive information (GitHub API tokens) from the event dict.

    Note: GH API tokens look different; this function is useless in reality.

    As these might occur at any depth, under any key, using regex is much simpler than
    any recursive approaches, looking at each key and value. It's much safer (harder to
    get wrong) and more performant (single pass, no recursion, stack cannot explode,
    ...).
    """

    if not isinstance(event_dict, str):
        # This check shouldn't be necessary when using type annotation to their full
        # potential. Used here to satisfy mypy for now.
        msg = f"Expected str, got {type(event_dict)}"
        raise TypeError(msg)

    # Clearly, `event_dict` is a misnomer by now...
    return re.sub(r"ghp_\d+", "***", event_dict)


def good() -> None:
    structlog.configure(
        processors=[
            TimeStamper(key="ts"),  # JSON -> JSON
            JSONRenderer(),  # JSON -> str
            scrub,  # str -> str
        ],
    )

    logger.info("hello", token="ghp_1234567890")


def woopsie() -> None:
    structlog.configure(
        processors=[
            TimeStamper(key="ts"),  # JSON -> JSON
            scrub,  # str -> str
            JSONRenderer(),  # JSON -> str
        ],
    )

    try:
        logger.info("hello", token="ghp_1234567890")
    except TypeError as e:
        print("This blows up, but `configure` passes type checks.")
        print(e)


good()

woopsie()
```

This is with structlog version 23.1.0, and **passes `mypy --strict`** (mypy version 1.4.1)!

The crux of the matter is the processor signature:

https://github.com/hynek/structlog/blob/6056aad1516346470a9c1485822daf531e30a1ba/src/structlog/typing.py#L63-L66

As far as I can see, the `Union` blows things wide open, with no way of saving it on the type level. You'd have to revert to runtime checks (`isinstance`, ...), which is dark ages Python :^) However, as can be seen in the above example, even then signatures would have to "lie", to some extent. The `scrub` function works, but contradicts the docs.

So I propose a stricter approach, because a single list of `Processor`s is not salvageable on the type level (it wouldn't even compile without bending backwards in most statically typed languages I reckon). They will always be reorderable invalidly, leading to runtime errors not caught statically. One approach is shown in the PR diff as it is right now: don't have a single `processors` list, but three:

- a list for (pre-)processors of `dict` -> `dict`
- a *single* one to go from `dict` -> `str`
- a list of (post-)processors of `str` -> `str`

### Areas of uncertainty

In the [docs](https://www.structlog.org/en/stable/processors.html#chains), it mentions:

> structlog only looks at the return value of the last processor. That means that as long as you control the next processor in the chain (i.e. the processor that will get your return value passed as an argument), you can return whatever you want.
>
> Returning a modified event dictionary from your processors is just a convention to make processors composable.

So I'm not sure whether I'm misusing the package by doing *string*, and not structured processing, myself. Second, it's worded such that a certain dynamic behavior is encouraged and wanted, which this PR would go against.

Also, I saw the signature of `configure` is 10 years old. Being at major version 23, I see you're not opposed to breaking changes (assuming semver), but it would be an API change in a long-running part.

(Also not sure about how well `WrappedLogger` plays into this. It being `Any` could throw a wrench into more ambitious typing endeavors; perhaps a `Protocol` could be used to bind wrapped loggers to at least *some* level of statically known behavior, but that's another topic.).

# Pull Request Check List

- [ ] **New APIs** are added to [`api.py`](https://github.com/hynek/structlog/blob/main/tests/typing/api.py).

### DIFF
--- a/example_misuse.py
+++ b/example_misuse.py
@@ -1,0 +1,70 @@
+import re
+
+import structlog
+
+from structlog.processors import JSONRenderer, TimeStamper
+from structlog.typing import EventDict, WrappedLogger
+
+
+def scrub(
+    logger: WrappedLogger,
+    name: str,
+    # ↓ annotation contradicts docs, and signature of `structlog.typing.Processor`, but
+    # works as intended
+    event_dict: EventDict | str,
+) -> str:
+    """Removes sensitive information (GitHub API tokens) from the event dict.
+
+    Note: GH API tokens look different; this function is useless in reality.
+
+    As these might occur at any depth, under any key, using regex is much simpler than
+    any recursive approaches, looking at each key and value. It's much safer (harder to
+    get wrong) and more performant (single pass, no recursion, stack cannot explode,
+    ...).
+    """
+
+    if not isinstance(event_dict, str):
+        # This check shouldn't be necessary when using type annotation to their full
+        # potential. Used here to satisfy mypy for now.
+        msg = f"Expected str, got {type(event_dict)}"
+        raise TypeError(msg)
+
+    # Clearly, `event_dict` is a misnomer by now...
+    return re.sub(r"ghp_\d+", "***", event_dict)
+
+
+def good() -> None:
+    structlog.configure(
+        processors=[
+            TimeStamper(key="ts"),  # JSON -> JSON
+            JSONRenderer(),  # JSON -> str
+            scrub,  # str -> str
+        ],
+    )
+
+    logger = structlog.get_logger()
+
+    logger.info("hello", token="ghp_1234567890")
+
+
+def woopsie() -> None:
+    structlog.configure(
+        processors=[
+            TimeStamper(key="ts"),  # JSON -> JSON
+            scrub,  # str -> str
+            JSONRenderer(),  # JSON -> str
+        ],
+    )
+
+    logger = structlog.get_logger()
+
+    try:
+        logger.info("hello", token="ghp_1234567890")
+    except TypeError as e:
+        print("This blows up, but `configure` passes type checks.")
+        print(e)
+
+
+good()
+
+woopsie()
--- a/src/structlog/_base.py
+++ b/src/structlog/_base.py
@@ -13,7 +13,7 @@ from typing import Any, Iterable, Mapping, Sequence
 
 from structlog.exceptions import DropEvent
 
-from .typing import BindableLogger, Context, PreProcessor, WrappedLogger
+from .typing import BindableLogger, Context, Processor, WrappedLogger
 
 
 class BoundLoggerBase:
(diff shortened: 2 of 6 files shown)


alexpovel:
Sorry CI servers 😞 hard at work for a load of invalid crap

hynek:
You're assuming wrong, structlog doesn't use SemVer but CalVer with [very strict backwards-compatibility guarantees](https://github.com/hynek/structlog/blob/main/.github/SECURITY.md).  structlog has effectively never broken compat so far. :)

That said, I feel your pain because I feel it every time I configure structlog myself, however your proposed changes would break too many people and kill one of the magics of structlog around its flexibility.

But. Given you only want type safety, why not just the following:

```python
structlog.configure(processors=type_safe(main=[…], render=[…], post=[…])
```

The beauty is that it would just return those lists concatenated and there wouldn't be any runtime impact. Assuming a better name. :)

alexpovel:
Woops, alright! So yeah, structlog's dynamicism and strict backward compatibility then make using typing to the fullest potential hard.

I might go with the `type_safe` approach in my code. Could perhaps be implemented as a [`TypeGuard`](https://peps.python.org/pep-0647/) of sorts, providing runtime safety as well. At the very least, any such approach will obsolete the many comments I put around `configure` and promote them to actual code!

hynek:
Feel free to share! Doesn’t have to be perfect/perfect PR. I’m curious too.

hynek:
I didn't mean within _this_ PR. ;) Please just open something whenever you have something to show!

PR #544, opened 2023-08-25
Proposal: change to _event avoid kw clash

mrenrich84:
# Summary
The current `meth` signature could create a clash with provided KW arguments.
In my case, I was trying to log out an AWS Lambda invocation event:
```py
logger.info("handle", event=event)
```
Which resulted in 
`TypeError: meth() got multiple values for argument 'event'`

The idea is to somehow avoid clashes, e.g. adding an underscore in front.

# Pull Request Check List
Haven't done any of the below yet.
I understand this might be a major change in the API, perhaps not something easy to achieve.
At this stage, I'm just letting maintainers know about this issue and brainstorm a possible solution.

- [ ] **New APIs** are added to [`api.py`](https://github.com/hynek/structlog/blob/main/tests/typing/api.py).

### DIFF
--- a/src/structlog/_log_levels.py
+++ b/src/structlog/_log_levels.py
@@ -160,7 +160,7 @@ def _make_filtering_bound_logger(min_level: int) -> type[FilteringBoundLogger]:
 
         name = _LEVEL_TO_NAME[level]
 
-        def meth(self: Any, event: str, *args: Any, **kw: Any) -> Any:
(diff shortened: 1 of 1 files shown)


hynek:
The name of event is sadly so ingrained in the ecosystem, we simply cannot break that.

Have you seen https://www.structlog.org/en/stable/recipes.html#renaming-the-event-key?

PR #552, opened 2023-09-12
blacken stdlib example from docs

wimglenn:
This is very minor, but if you copy/paste the configuration example from the docs ([Standard Library Logging](https://www.structlog.org/en/stable/standard-library.html#rendering-using-structlog-based-formatters-within-logging)) into your app, you'll get something with 3-space indents which [black](https://github.com/psf/black) wants to change.


hynek:
Ugh thanks! those 3-spaced RST files remain a constant frustration. :(

PR #556, opened 2023-09-28
Fix "default" TimeStamper args in docs/getting-started.md

NoamNol:
A small docs fix (:

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -54,7 +54,7 @@ structlog.configure(
         structlog.processors.add_log_level,
         structlog.processors.StackInfoRenderer(),
         structlog.dev.set_exc_info,
-        structlog.processors.TimeStamper(),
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -54,7 +54,7 @@ structlog.configure(
         structlog.processors.add_log_level,
         structlog.processors.StackInfoRenderer(),
         structlog.dev.set_exc_info,
-        structlog.processors.TimeStamper(),
+        structlog.processors.TimeStamper(fmt='%Y-%m-%d %H:%M:%S', utc=False),
```
hynek, line 57:
```suggestion
        structlog.processors.TimeStamper(fmt="%Y-%m-%d %H:%M:%S", utc=False),
```

hynek (changes requested):
Oops, thanks!

One tiny pedantic thin: we use double quotes everywhere for consistency. I realize it looks a bit silly to enforce, but especially the getting started should be "in perfect shape".

PR #565, opened 2023-10-20
callsite parameter fix while operating under `asyncio`

pahrohfit:
# Summary

Added support for proper callsite parameters while operating under `asyncio`

# Pull Request Check List

- [x] **New APIs** are added to [`api.py`](https://github.com/hynek/structlog/blob/main/tests/typing/api.py).

      The next version is the second number in the current release + 1. The first number represents the current year. So if the current version on PyPI is 23.1.0, the next version is gonna be 23.2.0. If the next version is the first in the new year, it'll be 24.1.0.

### DIFF
--- a/src/structlog/_frames.py
+++ b/src/structlog/_frames.py
@@ -5,16 +5,15 @@
 
 from __future__ import annotations
 
+import contextvars
 import sys
 import traceback
-
-import contextvars
 
 from io import StringIO
 from types import FrameType
 
-from .typing import ExcInfo
 from .contextvars import async_calling_stack
+from .typing import ExcInfo
 
 
 def _format_exception(exc_info: ExcInfo) -> str:
@@ -53,8 +52,13 @@ def _find_first_app_frame_and_name(
         tuple of (frame, name)
     """
     ctx = contextvars.copy_context()
+    ctx = contextvars.copy_context()
     ignores = ["structlog"] + (additional_ignores or [])
-    f = ctx.get(async_calling_stack) if async_calling_stack in ctx else sys._getframe()
+    f = (
+        ctx.get(async_calling_stack)
+        if async_calling_stack in ctx
+        else sys._getframe()
+    )
     name = f.f_globals.get("__name__") or "?"
     while any(tuple(name.startswith(i) for i in ignores)):
         if f.f_back is None:
--- a/src/structlog/_frames.py
+++ b/src/structlog/_frames.py
@@ -52,21 +52,20 @@ def _find_first_app_frame_and_name(
         tuple of (frame, name)
     """
     ctx = contextvars.copy_context()
-    ctx = contextvars.copy_context()
     ignores = ["structlog"] + (additional_ignores or [])
     f = (
         ctx.get(async_calling_stack)
         if async_calling_stack in ctx
         else sys._getframe()
     )
-    name = f.f_globals.get("__name__") or "?"
+    name = f.f_globals.get("__name__") or "?" # type: ignore[union-attr]
     while any(tuple(name.startswith(i) for i in ignores)):
-        if f.f_back is None:
+        if f.f_back is None: # type: ignore[union-attr]
             name = "?"
             break
-        f = f.f_back
+        f = f.f_back # type: ignore[union-attr]
         name = f.f_globals.get("__name__") or "?"
-    return f, name
+    return f, name  # type: ignore[return-value]
 
 
 def _format_stack(frame: FrameType) -> str:
--- a/src/structlog/_frames.py
+++ b/src/structlog/_frames.py
@@ -58,12 +58,12 @@ def _find_first_app_frame_and_name(
         if async_calling_stack in ctx
         else sys._getframe()
     )
-    name = f.f_globals.get("__name__") or "?" # type: ignore[union-attr]
+    name = f.f_globals.get("__name__") or "?"  # type: ignore[union-attr]
     while any(tuple(name.startswith(i) for i in ignores)):
-        if f.f_back is None: # type: ignore[union-attr]
+        if f.f_back is None:  # type: ignore[union-attr]
             name = "?"
             break
-        f = f.f_back # type: ignore[union-attr]
+        f = f.f_back  # type: ignore[union-attr]
         name = f.f_globals.get("__name__") or "?"
     return f, name  # type: ignore[return-value]
 
--- a/src/structlog/_frames.py
+++ b/src/structlog/_frames.py
@@ -5,7 +5,6 @@
 
 from __future__ import annotations
 
-import contextvars
 import sys
 import traceback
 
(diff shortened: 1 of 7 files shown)


pahrohfit:
Sample:

```python
import asyncio
import structlog

structlog.configure(
    processors=[
        structlog.processors.CallsiteParameterAdder(
            {
                structlog.processors.CallsiteParameter.FILENAME,
                structlog.processors.CallsiteParameter.FUNC_NAME,
                structlog.processors.CallsiteParameter.LINENO,
            },
        ),
    ],
    context_class=dict,
)


def sync_logging():
    logger.info("hello sync")

async def async_logging():
    await logger.ainfo("hello async via ainfo")

if __name__ == "__main__":
    sync_logging()

    loop = asyncio.get_event_loop()
    loop.run_until_complete(async_logging())
    loop.close()
```

Produces:
```
$ python structlog_test.py
{"event": "hello sync", "lineno": 22, "func_name": "sync_logging", "filename": "structlog_test.py"}
{"event": "hello async via ainfo", "lineno": 25, "func_name": "async_logging", "filename": "structlog_test.py"}
```

src/structlog/_frames.py
```diff
@@ -49,16 +51,21 @@ def _find_first_app_frame_and_name(
 
         tuple of (frame, name)
     """
+    ctx = contextvars.copy_context()
```
hynek, line 54: why the copy here if we only read?

src/structlog/_log_levels.py
```diff
@@ -91,16 +92,24 @@ def exception(
...
+    """
+    .. versionchanged:: 23.3.0
+       Implemented `contextvars.ContextVar` for holding and resetting async calling stack
+    """
     # Exception info has to be extracted this early, because it is no longer
     # available once control is passed to the executor.
     if kw.get("exc_info", True) is True:
         kw["exc_info"] = sys.exc_info()
 
+    _scs_token = async_calling_stack.set(sys._getframe().f_back)
```
hynek, line 104: There's no reason to have an underscored variable in a function, I think?

src/structlog/_log_levels.py
```diff
@@ -91,16 +92,24 @@ def exception(
...
     ctx = contextvars.copy_context()
-    return await asyncio.get_running_loop().run_in_executor(
+    _run = await asyncio.get_running_loop().run_in_executor(
         None,
         lambda: ctx.run(lambda: self.error(event, *args, **kw)),
     )
+    async_calling_stack.reset(_scs_token)
```
hynek, line 110: This – and the other locations with the same code – should be wrapped into a `try: run = ...; finally:  async_calling_stack.reset(scs_token)` to reset the token even if the logger throws an exception.

src/structlog/contextvars.py
```diff
@@ -30,6 +32,10 @@
 STRUCTLOG_KEY_PREFIX = "structlog_"
 STRUCTLOG_KEY_PREFIX_LEN = len(STRUCTLOG_KEY_PREFIX)
 
+async_calling_stack: contextvars.ContextVar[Any] = contextvars.ContextVar(
```
hynek, line 35: This variable should definitively be private and `[FrameType]` instead of `Any`, no?

src/structlog/_log_levels.py
```diff
@@ -91,16 +92,24 @@ def exception(
 async def aexception(
     self: FilteringBoundLogger, event: str, *args: Any, **kw: Any
 ) -> Any:
+    """
+    .. versionchanged:: 23.3.0
+       Implemented `contextvars.ContextVar` for holding and resetting async calling stack
```
hynek, line 97: This is an implementation detail. To the user the relevant part is "Callsite parameters are now also collected under asyncio." or something like that. In that vein: it really just needs an version changed on CallsiteParameters I would say.

hynek (changes requested):
Thanks this is cool! needs some work, but should be straight-forward.

src/structlog/_log_levels.py
pahrohfit: Makes sense -- updated the lines

src/structlog/_log_levels.py
pahrohfit: Truth -- cleaned up

src/structlog/_log_levels.py
pahrohfit: Didn't think about that, makes total sense.  Updated the blocks.

src/structlog/contextvars.py
pahrohfit: Yeah, I forgot to go back to it. Fixed.

src/structlog/_frames.py
pahrohfit: Balmers peak. Corrected and simplified.

pahrohfit:
How’s recent commit looking?

hynek:
Hm I tried to push changes into your branch but I'm getting a permission denied. Are you sure you haven't forgotten to grant the permission for that or am I just failing at git here?

src/structlog/contextvars.py
```diff
@@ -30,6 +33,10 @@
 STRUCTLOG_KEY_PREFIX = "structlog_"
 STRUCTLOG_KEY_PREFIX_LEN = len(STRUCTLOG_KEY_PREFIX)
 
+async_calling_stack: contextvars.ContextVar[
```
hynek, line 36: To make it analogue to our other contextvars:

```suggestion
_ASYNC_CALLING_STACK: contextvars.ContextVar[
```

Of course that requires to do the change all over the code base which I'm not gonna do in the web interface here. Feel free to do it locally with F2 in VS Code or something. ;)

tests/test_processors.py
```diff
@@ -850,12 +844,17 @@ def __init__(self):
...
+
         callsite_params = self.get_callsite_parameters()
         await logger.info("baz")
+        logger_params = json.loads(string_io.getvalue())
 
-        assert {"event": "baz", **callsite_params} == json.loads(
-            string_io.getvalue()
-        )
+        [callsite_params.pop(key) for key in exclude_keys]
+        [logger_params.pop(key) for key in exclude_keys]
```
hynek, line 855: That seems unnecessarily clever:

```suggestion
        # These are different when running under async
        for key in ["thread", "thread_name"]:
            callsite_params.pop(key)
            logger_params.pop(key)
```

tests/test_processors.py
```diff
@@ -850,12 +844,17 @@ def __init__(self):
 
         logger = structlog.stdlib.get_logger()
 
+        # These are different when running under async
+        exclude_keys = ["thread", "thread_name"]
+
```
hynek, line 849: This can be removed I think (see below).

hynek (changes requested):

pahrohfit:
> Hm I tried to push changes into your branch but I'm getting a permission denied. Are you sure you haven't forgotten to grant the permission for that or am I just failing at git here?

image

It shows the relevant box checked ...

src/structlog/contextvars.py
pahrohfit: hah, yeah, just made the search/replace

tests/test_processors.py
pahrohfit: Thought it was neater if that list ever had to grow.

tests/test_processors.py
pahrohfit: Yeap, killed with the list context pop change.

hynek:
hm no if the checkmark is set, it should work. maybe it's because you opened your PR from your own main branch which has protections. You should avoid doing that in general, none-the-least because once this PR is merged, you'll have a bad time syncing your fork due to squash and merge.

sadly that's an information that comes too late for the PR checklist ;)

hynek (approved):
Thanks!

PR #567, opened 2023-10-30
`AddCallingClassPath` processor added

### LINKED ISSUE #386: Add class name support for CallsiteParameterAdder
It may be nice to include support for class name in `CallsiteParameterAdder`, not sure how practical/easy it will be though, or how reliable it is to get the class name of the callsite.

Original request from @gzmarstone here: https://github.com/hynek/structlog/pull/380#issuecomment-995873510

> I know its closed, but what about class if available as well... use case would be to find all of the logging calls in a particular class... You can check if the first arg is self, if it is then get the **class** and the **name** of that? I used this example to log the classname via structlog: https://stackoverflow.com/questions/2203424/python-how-to-retrieve-class-information-from-a-frame-object

### LINKED ISSUE #492: Adding ability to print full namespace of module for CallsiteParameterAdder
When using `CallsiteParameter.MODULE` in `CallsiteParameterAdder`, it's a bit too vague for my tastes - for example, a module named `config` has no meaning unless I know what package it's referring to. For example, I may have a module called `foo.bar.config` and another package could have a module called `baz.buz.config`, and it would not be possible to tell the difference with that value currently.

I know that there's `CallsiteParameter.PATHNAME`, but that's overly verbose - although I guess I could transform it to be something more palatable... At the end of the day though, it would be nice if there was a `CallsiteParameter` that would just print out `foo.bar.config`.

pahrohfit:
# Summary

Added a `class_path` event entry processor as `AddCallingClassPath` (fixes #386, fixes #492). 

Takes an optional `set` or `list` to limit which logging levels to add this at, as lookups potentially can add overhead (though should run in N(1)).

# Pull Request Check List

- [X] Do **not** open pull requests from your `main` branch – **use a separate branch**!

  There's a ton of footguns waiting if you don't heed this warning. You can still go back to your project, create a branch from your main branch, push it, and open the pull request from the new branch.

  This is not a pre-requisite for your your pull request to be accepted, but **you have been warned**.
- [X] **New APIs** are added to our typing tests in [`api.py`](https://github.com/hynek/structlog/blob/main/tests/typing/api.py).
    - [X] Changed/added classes/methods/functions have appropriate `versionadded`, `versionchanged`, or `deprecated` [directives](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded).

      The next version is the second number in the current release + 1. The first number represents the current year. So if the current version on PyPI is 23.1.0, the next version is gonna be 23.2.0. If the next version is the first in the new year, it'll be 24.1.0.
- [X] Documentation in `.rst` and `.md` files is written using [**semantic newlines**](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).
- [X] Changes (and possible deprecations) are documented in the [**changelog**](https://github.com/hynek/structlog/blob/main/CHANGELOG.md).
- [X] Consider granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), so maintainers can fix minor issues themselves without pestering you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ You can find our backwards-compatibility policy [here](https://github.com/hynek/
 
 ## [Unreleased](https://github.com/hynek/structlog/compare/23.2.0...HEAD)
 
+### Added
+
+- `structlog.processors.AddCallingClassPath()` added, which will attempt to determine the calling class path and add it as `class_path` to the event dict. Takes an optional `levels` `dict`|`set` to limit which `logging.LEVEL` to include the addition in.
+
 ### Fixed
 
 - `AsyncBoundLogger` now properly supports callsite parameters.
@@ -23,7 +27,6 @@ You can find our backwards-compatibility policy [here](https://github.com/hynek/
   [#561](https://github.com/hynek/structlog/issues/561)
 
 - `structlog.threadlocal.tmp_bind()` now also works `BoundLoggerLazyProxy` (in other words: before anything is bound to a bound logger).
-
 
 ## [23.2.0](https://github.com/hynek/structlog/compare/23.1.0...23.2.0) - 2023-10-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ You can find our backwards-compatibility policy [here](https://github.com/hynek/
 
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


hynek:
Welcome to the "Why always open PRs from branches" lesson we all have to learn. :( Sorry for the pain.

PR #604, opened 2024-03-13
structlog.testing.CapturingLogger now handles bind

CarliJoy:
# Summary

Using the `CapturingLogger` for a function that does `logger.bind` currently fails 

# Pull Request Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ You can find our backwards-compatibility policy [here](https://github.com/hynek/
 - `structlog.processors.LogfmtRenderer` now escapes backslashes and double quotes.
   [#594](https://github.com/hynek/structlog/pull/594)
 
+- `structlog.testing.CapturingLogger` now handles `bind`, `unbind` and `new` as expected by `BoundLogger` returning itself.
+  [#604](https://github.com/hynek/structlog/pull/604)
 
 
 ## [24.1.0](https://github.com/hynek/structlog/compare/23.3.0...24.1.0) - 2024-01-08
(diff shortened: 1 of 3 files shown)


hynek:
I think there might be a bit of a confusion, or I am profoundly confused by my own project. 😵‍💫

CapturingLogger is *not* a bound logger and shouldn't ever appear anywhere where a `bind` call should happen. It should be used for [logger factories](https://www.structlog.org/en/stable/configuration.html#logger-factories) that receive that final event dict.

How did you run into the situation that you needed bind et al?

CarliJoy:
Oh I see.

I only looked into the source code when using but somehow missed the example/usage from the docs:

```python
>>> import structlog
>>> 
>>> cf = structlog.testing.CapturingLoggerFactory()
>>> structlog.configure(logger_factory=cf, processors=[structlog.processors.JSONRenderer()])
>>> log = get_logger()
>>> log.info("test!")
>>> cf.logger.calls
[CapturedCall(method_name='info', args=('{"event": "test!"}',), kwargs={})]
````

I simply did 

```python
from structlog.testing import CapturingLogger, CapturedCall


def my_func(logger):
    do()
    logger.info("done")

def log_call(__method: str, *args: Any, **kwargs: Any) -> CapturedCall:
    """
    Nicer way to create CapturedCall
    """
    return CapturedCall(__method, args, kwargs)

def test_my_func():
     logger = CapturingLogger()
     my_func(logger)
     assert logger.calls == [log_call("info", "done")]

```

That looked simple enough for me to not look into the documentation to verify.

hynek:
Right, I’m gonna close this now that this is resolved. :)

kieran-sf:
I am running into a similar problem. I am trying to mock my logger and then check the captured logs for a log message but I get an error when trying to use the CaptureLogger because it isn't a BinableLogger.

PR #607, opened 2024-03-27
optimize `_find_first_app_frame_and_name`

methane:
# Summary

Skip temporary tuple.

Performance impact is very small. But it makes code simpler.

# Pull Request Check List

### DIFF
--- a/src/structlog/_frames.py
+++ b/src/structlog/_frames.py
@@ -58,7 +58,7 @@ def _find_first_app_frame_and_name(
     ignores = ["structlog"] + (additional_ignores or [])
     f = _ASYNC_CALLING_STACK.get(_getframe())
     name = f.f_globals.get("__name__") or "?"
-    while any(tuple(name.startswith(i) for i in ignores)):
+    while any(name.startswith(i) for i in ignores):
         if f.f_back is None:
             name = "?"
             break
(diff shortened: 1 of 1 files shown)


methane:
Since peformance difference is almost invisible to users, I am not sure this need changelog.

hynek:
Interesting, I could swear it's necessary! Could it be this changed at some point? Maybe 2 → 3?

hynek:
ah wait, found it: #174

I don't think the problem it's solving is gone? We should probably add a comment to it.

methane:
Oh, it's really strange bug.
I am trying to reproduce but I can not yet...

Anyway, list comprehension is faster and efficient than `tuple(genexpr)`. tuple constructor need to build temporary list to know length of genexpr.

```
$ python -m timeit -s 's="abcde"' -- '[c for c in s]'
5000000 loops, best of 5: 76.8 nsec per loop

$ python -m timeit -s 's="abcde"' -- 'tuple(c for c in s)'
1000000 loops, best of 5: 247 nsec per loop
```

methane:
Just for fun:

```py
    while [i for i in ignores if name.startswith(i)]:
```

```
$ .venv/bin/python logging_bench.py # while [i for i in...]
<_FixedFindCallerLogger hello (INFO)>
8.033us
```

This is the fastest one-liner.

methane:
I forgot that str.startswith() takes a tuple. It is simple, fast, and safe.

```
$ .venv/bin/python logging_bench.py
<_FixedFindCallerLogger hello (INFO)>
6.974us
```

hynek:
Nice, thanks!

Issue #634, opened 2024-07-23
ExceptionRenderer w/ ExceptionDictTransformer unable to handle exception log outside exception handler


raqbit:
When using the `exception` log method outside an exception handler, the `ExceptionDictTransformer` raises an `AttributeError` as it calls `extract`, which does not handle the case where the given exception type is `None`.

While according to the Python docs[^1] this method should only be used from exception handlers, I ran into this issue as the aio-pika (aiormq) library uses it outside an exception handler ([issue](https://github.com/mosquito/aiormq/issues/200)). (I have configured structlog as the formatter for all loggers, using [this](https://www.structlog.org/en/stable/standard-library.html#rendering-using-structlog-based-formatters-within-logging) setup)

Note that the default `_format_exception` formatter handles this case properly, as it handles `(None, None, None)` exc_info[^2].

[^1]: https://docs.python.org/3/library/logging.html#logging.Logger.exception
[^2]: https://github.com/hynek/structlog/blob/e48a553979b00ab580d8ab0e55b5cf96e752a324/src/structlog/_frames.py#L25
---

Reproduction:
```
import structlog
print(structlog.__version__) # 24.4.0
structlog.configure([structlog.processors.dict_tracebacks])
structlog.get_logger().exception("oh no") # AttributeError: 'NoneType' object has no attribute '__name__'. Did you mean: '__ne__'?
```

hynek:
Is this a new issue?

@sscherfke I'm afraid there's work for you. 🤓

raqbit:
Can reproduce in 24.2.0, which does not include #627
```
Python 3.11.9 (main, Apr  6 2024, 17:59:24) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import structlog
>>> print(structlog.__version__)
24.2.0
>>> structlog.configure([structlog.processors.dict_tracebacks])
>>> structlog.get_logger().exception("oh no")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<snip>/venv/lib/python3.11/site-packages/structlog/_native.py", line 45, in exception
    return self.error(event, *args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<snip>/venv/lib/python3.11/site-packages/structlog/_native.py", line 134, in meth
    return self._proxy_to_logger(name, event, **kw)
[...]

PR #635, opened 2024-07-24
WIP: Handle Logger.exception() outside "except" block

sscherfke:
Fixes: #634

# Summary

# Pull Request Check List

- [ ] Do **not** open pull requests from your `main` branch – **use a separate branch**!

### DIFF
--- a/src/structlog/processors.py
+++ b/src/structlog/processors.py
@@ -19,7 +19,7 @@ import sys
 import threading
 import time
 
-from types import FrameType
+from types import FrameType, TracebackType
 from typing import (
     Any,
     Callable,
@@ -39,7 +39,12 @@ from ._frames import (
 from ._log_levels import NAME_TO_LEVEL, add_log_level
 from ._utils import get_processname
 from .tracebacks import ExceptionDictTransformer
-from .typing import EventDict, ExceptionTransformer, ExcInfo, WrappedLogger
+from .typing import (
+    EventDict,
+    ExceptionTransformer,
+    ExcInfo,
+    WrappedLogger,
+)
 
 
 __all__ = [
@@ -409,10 +414,8 @@ class ExceptionRenderer:
         self, logger: WrappedLogger, name: str, event_dict: EventDict
     ) -> EventDict:
         exc_info = _figure_out_exc_info(event_dict.pop("exc_info", None))
-        if exc_info and exc_info != (None, None, None):
-            event_dict["exception"] = self.format_exception(
-                cast(ExcInfo, exc_info)
-            )
+        if exc_info:
+            event_dict["exception"] = self.format_exception(exc_info)
 
         return event_dict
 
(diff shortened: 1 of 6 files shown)


src/structlog/processors.py
```diff
@@ -586,7 +586,7 @@ def __call__(
         return event_dict
 
 
-def _figure_out_exc_info(v: Any) -> ExcInfo:
+def _figure_out_exc_info(v: Any) -> Union[ExcInfo, Tuple[None, None, None]]:
```
sscherfke, line 589: The commit hook wants me to write `ExcInfo | tuple[None, None, None]` but this will break on older Python versions.  For Typed Settings, I configured

```toml
[tool.ruff]
target-version = "py38"
```

to fix this.

hynek: this shouldn't break with `from __future__ import annotations`

src/structlog/processors.py
```diff
@@ -642,7 +643,7 @@ def __call__(
         exc = event_dict.pop("exception", None)
         if exc is None:
             exc_info = _figure_out_exc_info(event_dict.pop("exc_info", None))
-            if exc_info:
+            if exc_info != (None, None, None):
```
hynek, line 646: looks like this was the actual bug, right? because a negative result from _figure_out_exc_info isn't falsey

src/structlog/processors.py
```diff
@@ -407,11 +409,9 @@ def __init__(
     def __call__(
         self, logger: WrappedLogger, name: str, event_dict: EventDict
     ) -> EventDict:
-        exc_info = event_dict.pop("exc_info", None)
-        if exc_info:
-            event_dict["exception"] = self.format_exception(
-                _figure_out_exc_info(exc_info)
-            )
+        exc_info = _figure_out_exc_info(event_dict.pop("exc_info", None))
+        if exc_info != (None, None, None):
```
raqbit, line 413: It seems like this would pass a `None` exc_info in the event dict into the formatter. Is this expected?

src/structlog/processors.py
sscherfke: Yes, more or less. The `(None, None, None)` case was handled elsewhere (surprisingly, from a typing perspective ;-)), but it only worked for the default exception formatter.

src/structlog/processors.py
sscherfke: Changed it to `if exc_info and exc_info != (None, None, None)`. This should work.

Not happy though, that `exc_info` can be literally anything.  I will take a look if this can somehow be changed/improved.

tests/processors/test_renderers.py
```diff
@@ -580,6 +580,4 @@ def test_no_exception(self, ei):
         """
         A missing exception does not blow up.
         """
-        assert {"exception": "MISSING"} == format_exc_info(
-            None, None, {"exc_info": ei}
-        )
+        assert {} == format_exc_info(None, None, {"exc_info": ei})
```
sscherfke, line 583: Is this okay for you, @hynek ?  The `format_exc_info()` handled the `(None, None, None)` (even though its signature suggested that `exc_info: ExcInfo`) and returned `{"excepiton": "MISSING"}`.

Now, `(None, None, None)` will no longer be passed to it and thus there will no longer be a `MISSING`.

sscherfke:
I think that 0c02f29fa259c3173ed0c0b6247a093f41ac312f _really_ fixes the problem in the right way. I added a longer description to the commit to explain why.

hynek:
close on the wish of the author due to lack of feedback from OP

tests/processors/test_renderers.py
sscherfke: @hynek ping

hynek:
CleanShot 2024-11-28 at 06 48 19@2x


looks like I can't reopen it – you'll have to recreate a new one. never seen this before

PR #636, opened 2024-07-25
Tweak `add_log_level` type

Tinche:
This doesn't actually need to be a `logging.Logger`, right?


hynek (approved):
naw it doesn't. it used to be a stdlib processor living in the stdlib namespace but that's been a while

PR #658, opened 2024-10-03
Switch base methods to return Self instead of BoundLoggerBase

aThorp96:
# Summary

This Pull Request switches several `BoundLoggerBase` methods to annotate a return type of `typing_extensions.Self` instead of `BoundLoggerBase`. This change is consistent with the code which [returns specifically](https://github.com/hynek/structlog/blob/main/src/structlog/_base.py#L70) `self.__class__()` and with the tests that ensure the [type is preserved when `.bind()` is called](https://github.com/hynek/structlog/blob/main/tests/test_base.py#L82).

This should not have any runtime changes, but instead fixes a typing bug which caused the following code to incorrectly fail type-checking:
```python
from structlog import get_logger, BoundLogger

log: BoundLogger = get_logger()
log.info("This is OK")
log = log.bind() #  error: Type "BoundLoggerBase" is not assignable to declared type "BoundLogger"
log.info("This should also be okay")
```

Before this change:
Screenshot 2024-10-03 at 4 48 17 PM

After this change:
Screenshot 2024-10-03 at 4 48 54 PM

# Pull Request Check List
It does not appear to me that any of these apply, but I am happy to make any necessary changes to the PR or do any upkeep tasks 

- [ ] Do **not** open pull requests from your `main` branch – **use a separate branch**!

### DIFF
--- a/src/structlog/_base.py
+++ b/src/structlog/_base.py
@@ -9,7 +9,8 @@ Logger wrapper and helper class.
 
 from __future__ import annotations
 
-from typing import Any, Iterable, Mapping, Sequence
+from typing import Any, Iterable, Mapping, Sequence, Self
+from typing_extensions import Self
 
 from structlog.exceptions import DropEvent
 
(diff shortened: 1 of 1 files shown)


aThorp96:
Forgot to switch PR from my main branch to my non-main branch

PR #659, opened 2024-10-03
Switch base methods to return Self instead of BoundLoggerBase

aThorp96:
# Summary

This Pull Request switches several `BoundLoggerBase` methods to annotate a return type of `typing_extensions.Self` instead of `BoundLoggerBase`. This change is consistent with the code which [returns specifically](https://github.com/hynek/structlog/blob/main/src/structlog/_base.py#L70) `self.__class__()` and with the tests that ensure the [type is preserved when `.bind()` is called](https://github.com/hynek/structlog/blob/main/tests/test_base.py#L82).

This should not have any runtime changes, but instead fixes a typing bug which caused the following code to incorrectly fail type-checking:
```python
from structlog import get_logger, BoundLogger

log: BoundLogger = get_logger()
log.info("This is OK")
log = log.bind() #  error: Type "BoundLoggerBase" is not assignable to declared type "BoundLogger"
log.info("This should also be okay")
```

Before this change:
Screenshot 2024-10-03 at 4 48 17 PM

After this change:
Screenshot 2024-10-03 at 4 48 54 PM

# Pull Request Check List
It does not appear to me that any of these apply, but I am happy to make any necessary changes to the PR or do any upkeep tasks 

### DIFF
--- a/src/structlog/_base.py
+++ b/src/structlog/_base.py
@@ -13,7 +13,7 @@ from typing import Any, Iterable, Mapping, Sequence
 
 from structlog.exceptions import DropEvent
 
-from .typing import BindableLogger, Context, Processor, WrappedLogger
+from .typing import BindableLogger, Context, Processor, WrappedLogger, Self
 
 
 class BoundLoggerBase:
@@ -62,7 +62,7 @@ class BoundLoggerBase:
     def __ne__(self, other: object) -> bool:
         return not self.__eq__(other)
 
-    def bind(self, **new_values: Any) -> BoundLoggerBase:
+    def bind(self, **new_values: Any) -> Self:
         """
         Return a new logger with *new_values* added to the existing ones.
         """
@@ -72,7 +72,7 @@ class BoundLoggerBase:
             self._context.__class__(self._context, **new_values),
         )
 
-    def unbind(self, *keys: str) -> BoundLoggerBase:
+    def unbind(self, *keys: str) -> Self:
         """
         Return a new logger with *keys* removed from the context.
 
@@ -85,7 +85,7 @@ class BoundLoggerBase:
 
         return bl
 
-    def try_unbind(self, *keys: str) -> BoundLoggerBase:
+    def try_unbind(self, *keys: str) -> Self:
         """
         Like :meth:`unbind`, but best effort: missing keys are ignored.
 
@@ -97,7 +97,7 @@ class BoundLoggerBase:
 
         return bl
 
-    def new(self, **new_values: Any) -> BoundLoggerBase:
+    def new(self, **new_values: Any) -> Self:
         """
         Clear context and binds *new_values* using `bind`.
 
--- a/src/structlog/_base.py
+++ b/src/structlog/_base.py
@@ -9,11 +9,19 @@ Logger wrapper and helper class.
 
 from __future__ import annotations
 
+import sys
+
 from typing import Any, Iterable, Mapping, Sequence
 
 from structlog.exceptions import DropEvent
 
-from .typing import BindableLogger, Context, Processor, Self, WrappedLogger
+from .typing import BindableLogger, Context, Processor, WrappedLogger
+
+
+if sys.version_info >= (3, 11):
+    from typing import Self as Self
+else:
+    from typing_extensions import Self as Self
 
 
 class BoundLoggerBase:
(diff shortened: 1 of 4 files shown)


aThorp96:
All of the CI here is failing because it cannot import `typing_extensions`. Not sure what's going on with that; `typing_extensions` was already a dependency...

hynek:
typing extensions needs to be a conditional import as done in https://github.com/hynek/structlog/pull/642 – and speaking of it, can you check please if your problems aren't fixed by that already?

If not, feel free to import `Self` from `structlog.typing`.

aThorp96:
Thanks @hynek! I missed that `typing_extensions.Self` was only available on python <3.11. It makes sense to only import it for those python versions. I've switched to `structlog.typing.Self` since it already handles the version-specific cases. Not sure why the PR isn't updating yet with the changes I've pushed.

Unfortunately #642 does not fix the type error I am seeing, since `BoundLogger` directly sublcasses `BoundLoggerBase`. The screenshot of the Pyright error is on `main` yesterday, on  9e9711c specifically.

hynek:
In that case, thanks for catching it before shipping #642. :) Is there maybe a way to add a typing test to https://github.com/hynek/structlog/blob/main/tests/typing/api.py that would've caught it?

Either way, you haven't pushed you changes yet. 🤓

aThorp96:
I think github is being odd; [aThorp96:switch-error-type-narrowing](https://github.com/aThorp96/structlog/commits/switch-error-type-narrowing/) is up to date with the change. I rebased the branch on my end to squash the changes, and I think it messed w/ Github's remote ref to my branch.... I'll try to add a test to the API and open a new PR against the branch, since you can't change the PR's HEAD ref

tests/typing/api.py
```diff
@@ -323,6 +323,19 @@ async def typecheck_stdlib_async() -> None:
     await logger.alog(logging.CRITICAL, "async log")
 
 
+def typecheck_bound_logger_return() -> None:
+    blogger: structlog.BoundLogger = structlog.get_logger(__name__)
```
aThorp96, line 327: Interesting find here, and perhaps the reason that this hasn't come up sooner: 
This test would fail with `structlog.BoundLogger` before this PR, but it would pass with `structlog.stdlib.BoundLogger`.

aThorp96: Mmm, after hitting this, I think this is actually the root of the issue I'm experiencing; I don't think I realized that in some places I was working with a `structlog.BoundLogger` and other places I was working with a `structlog.stdlib.BoundLogger`, and if I did, I did not expect there to be a difference. My assumption would have been that the former was a top-level export of the latter. At least then I think my type error woes are gone. I'll still get this PR over the finish line, since i think it's correct and valuable.

src/structlog/_base.py
```diff
@@ -9,13 +9,21 @@
...
+import sys
+
 from typing import Any, Iterable, Mapping, Sequence
 
 from structlog.exceptions import DropEvent
 
 from .typing import BindableLogger, Context, Processor, WrappedLogger
 
 
+if sys.version_info >= (3, 11):
```
hynek, line 21: Any reason why you’re doing this guarded import dance instead of adding Self to the `.typing` import three lines up?

aThorp96: I initially tried this in [c04f3ec8](https://github.com/hynek/structlog/pull/659/commits/c04f3ec824e768933fc3c91f86a179557276fa2b#diff-7f889fd0855ea2564d56eaefdf3783b06b62a9d9585281cdebefcd884f027ca2R16), but pre-commit has [a rule](https://github.com/hynek/structlog/actions/runs/11188922225/job/31108545798?pr=659#step:6:29) against importing an attribute that eh module doesn't explicitly export. I tried to get around this a couple times ([fa43344b](https://github.com/hynek/structlog/pull/659/commits/fa43344ba466774afd39464c5a991be96033056f), [9206ef49](https://github.com/hynek/structlog/pull/659/commits/9206ef49bbefed3044aa6d35b387db520b0bef9e)) but I couldn't find a way which worked without violating one pre-commit rule or another. If there's a way to explicitly re-export `typing.Self` from `structlog.typing`, I think it'd be preferable to import it from there.

hynek: ugh right thanks. Maybe I'll add a _compat module one day, like in attrs. one can re-export using `__all__`

hynek (approved):
Thanks!

PR #670, opened 2024-11-09
Use real methods for fatal/warn

### LINKED ISSUE #648: use real method for `warn` and `fatal` to improve UX
`warn` method and `fatal` method of `BoundLogger` are created with `warn = warning`, make both pycharm and vscode not automatically add `(` after auto-complete, write a real method like `def warn(self, event: str | None = None, *args: Any, **kw: Any) -> Any:` will help this.

This is not a big issue but it would be nice to improve UX.

hynek:
For better DX in IDEs.

Fixes #648

like this @trim21 or is something missing?


trim21:
lgtm, thanks

PR #675, opened 2024-11-13
Make sure locals_max_{length,string} work as expected in case they are None

### LINKED ISSUE #672: TypeError when setting ExceptionDictTransformer's constructor arguments locals_max_string and locals_max_length to None
As of version 24.3.0 (see #627), `ExceptionDictTransformer` interprets the constructor arguments `locals_max_string` and `locals_max_length` as type `int | None` where `None` supposedly disables abbreviation or truncation. However, if you set one of those to `None`, you get a `TypeError`:

```
TypeError: '<' not supported between instances of 'NoneType' and 'int'
``` 

This is because of the following snippet:

https://github.com/hynek/structlog/blob/51641b764085c1b907a9b57857c1e71b2810a70e/src/structlog/tracebacks.py#L385-L390

It should contain logic that handles the `None` case.

Sh4pe:
This PR fixes the case where truncation or abbreviation in `ExceptionDictTransformer` is disabled using `locals_max_length` or `locals_max_string`.

Closes #672.


Sh4pe:
Uhm, what about the pre-commit check? What is wrong there?

tests/test_tracebacks.py
```diff
@@ -531,6 +531,40 @@ def bar(n):
...
+        ),
+        (
+            {"locals_max_length": None},
+            list(range(tracebacks.LOCALS_MAX_LENGTH + 10)),
+        ),
+    ],
+)
+def test_exception_dict_transformer_locals_max_accept_none_argument(
+    kwargs, local_variable
+):
```
hynek, line 549: Sorry for the late response! The CI is failing because your test doesn't have a docstring in the form of https://jml.io/test-docstrings/

See also: https://hynek.me/articles/document-your-tests/

Sh4pe: Alright, got it, thanks. I tried to add a suitable docstring, hope that suffices.

tests/test_tracebacks.py
```diff
@@ -531,6 +531,45 @@ def bar(n):
...
+        )
+        result = format_json((type(e), e, e.__traceback__))
+
+    _ = my_local
+
+    assert len(result) == 1
+    assert len(result[0]["frames"]) == 1
+
+    frame = result[0]["frames"][0]
+    assert frame["locals"]["my_local"].strip("'") == str(local_variable)
```
hynek, line 570:
```suggestion
    frame = result[0]["frames"][0]

    assert frame["locals"]["my_local"].strip("'") == str(local_variable)
```

hynek:
thanks!

PR #680, opened 2024-11-28
Handle Logger.exception() outside "except" block

### LINKED ISSUE #634: ExceptionRenderer w/ ExceptionDictTransformer unable to handle exception log outside exception handler

sscherfke:
# Summary

Fixes #634, follow-up of #635

# Pull Request Check List


hynek:
the tests are red 😇

also have you seen, that there's been changes to missing tracebacks lately – notably #657? 🤔 not sure if that maybe interacts somehow with your changes since it's thematically adjacent.

sscherfke:
Didn't have time yet to fix the tests.  Just recreated the MR for now. #657 is superfluous now. The problem it handled should no longer be possible with this PR.

sscherfke:
Check's are green now.  You should still carefully review the changes. ;-)

sscherfke:
Do you need anything else from me for this PR to get merged? :-)

hynek:
Sorry for the delays, I’ll get to structlog (there‘s quite a bit of open issues around exceptions 😰) right after pushing out attrs. I _think_ it’s ready.

tests/test_tracebacks.py
```diff
@@ -732,12 +734,72 @@ def test_json_traceback_value_error(
...
+class TestLogException:
     """
-    ExceptionDictTransformer returns an empty list if exc_info is missing.
+    Higher level "integration tests" for "Logger.exception()".
     """
-    transformer = tracebacks.ExceptionDictTransformer()
 
-    result = transformer(exc_info=(None, None, None))
+    @pytest.fixture
+    def cap_logs(self) -> structlog.testing.LogCapture:
```
hynek, line 743: you can make these static since you don't use self

tests/test_tracebacks.py
```diff
@@ -732,12 +734,72 @@ def test_json_traceback_value_error(
...
+        self, logger: structlog.Logger, cap_logs: structlog.testing.LogCapture
+    ) -> None:
+        """
+        "Logger.exception()" should not be called outside an "except" block
+        but this cases is gracefully handled and does not lead to an exception.
+
+        See: https://github.com/hynek/structlog/issues/634
+        """
+        logger.exception("onoes")
+        assert [{"event": "onoes", "log_level": "error"}] == cap_logs.entries
```
hynek, line 805:
```suggestion
        logger.exception("onoes")

        assert [{"event": "onoes", "log_level": "error"}] == cap_logs.entries
```

tests/test_tracebacks.py
```diff
@@ -732,12 +734,72 @@ def test_json_traceback_value_error(
...
+        The log row contains a traceback when "Logger.exception()" is
+        explicitly called with an exception instance.
+        """
+        try:
+            1 / 0
+        except ZeroDivisionError as e:
+            logger.exception("onoes", exception=e)
+
+        log_row = cap_logs.entries[0]
+        assert log_row["exception"][0]["exc_type"] == "ZeroDivisionError"
```
hynek, line 778:
```suggestion
        log_row = cap_logs.entries[0]

        assert log_row["exception"][0]["exc_type"] == "ZeroDivisionError"
```

tests/test_tracebacks.py
hynek: So, what this whole thing does is avoiding an empty/`MISSING` `exception` key altogether if there is no exception? I wonder how breaking this would be considered? 🤔

tests/test_tracebacks.py
```diff
@@ -732,12 +734,72 @@ def test_json_traceback_value_error(
...
+        The log row contains a traceback when "Logger.exception()" is called
+        in an "except" block but without explicitly passing the exception.
+        """
+        try:
+            1 / 0
+        except ZeroDivisionError:
+            logger.exception("onoes")
+
+        log_row = cap_logs.entries[0]
+        assert log_row["exception"][0]["exc_type"] == "ZeroDivisionError"
```
hynek, line 793:
```suggestion
        log_row = cap_logs.entries[0]

        assert log_row["exception"][0]["exc_type"] == "ZeroDivisionError"
```

tests/test_tracebacks.py
```diff
@@ -732,12 +734,72 @@ def test_json_traceback_value_error(
         tracebacks.ExceptionDictTransformer(**kwargs)
 
 
-def test_exception_dict_transformer_missing_exc_info():
+class TestLogException:
     """
-    ExceptionDictTransformer returns an empty list if exc_info is missing.
+    Higher level "integration tests" for "Logger.exception()".
```
hynek, line 739:
```suggestion
    Higher level integration tests for `Logger.exception()`.
```

tests/test_tracebacks.py
```diff
@@ -732,12 +734,72 @@ def test_json_traceback_value_error(
...
+        try:
+            yield logger
+        finally:
+            structlog.configure(processors=old_processors)
+
+    def test_log_explicit_exception(
+        self, logger: structlog.Logger, cap_logs: structlog.testing.LogCapture
+    ) -> None:
+        """
+        The log row contains a traceback when "Logger.exception()" is
```
hynek, line 769:
```suggestion
        The log row contains a traceback when `Logger.exception()` is
```

tests/test_tracebacks.py
```diff
@@ -732,12 +734,72 @@ def test_json_traceback_value_error(
...
+
+        log_row = cap_logs.entries[0]
+        assert log_row["exception"][0]["exc_type"] == "ZeroDivisionError"
+
+    def test_log_implicit_exception(
+        self, logger: structlog.Logger, cap_logs: structlog.testing.LogCapture
+    ) -> None:
+        """
+        The log row contains a traceback when "Logger.exception()" is called
+        in an "except" block but without explicitly passing the exception.
```
hynek, line 785:
```suggestion
        The log row contains a traceback when `Logger.exception()` is called
        in an `except` block but without explicitly passing the exception.
```

tests/test_tracebacks.py
```diff
@@ -732,12 +734,72 @@ def test_json_traceback_value_error(
...
+            logger.exception("onoes")
+
+        log_row = cap_logs.entries[0]
+        assert log_row["exception"][0]["exc_type"] == "ZeroDivisionError"
+
+    def test_no_exception(
+        self, logger: structlog.Logger, cap_logs: structlog.testing.LogCapture
+    ) -> None:
+        """
+        "Logger.exception()" should not be called outside an "except" block
```
hynek, line 799:
```suggestion
        `Logger.exception()` should not be called outside an `except` block
```

hynek (changes requested):
This very much needs a changelog entry – and as I asked inline, I wonder how breaking this would be considered given how many complains we got for the last release. 🤔

tests/test_tracebacks.py
sscherfke: Yes, no exception => no `exception` key in the event dict.  For me, this is the "obvious" expected behavior but I guess, you could also argue otherwise.

This change breaks other peoples integration tests if they test that logging an exception when there is no exception produces `"exception": "MISSING"`.

But from a "production perspective", I would not classify this change as "breaking". But again, you could argue otherwise.

sscherfke:
I'll add a changelog entry when the code is ready.

tests/test_tracebacks.py
sscherfke: No. This breaks pytest on older Python versions (`AttributeError: 'staticmethod' object has no attribute '__name__'`) and fixing _that_ is more cumbersome than living with a non-static method not using self. :-)

hynek (approved):
ok changelog pls and let's get this behind us

sscherfke:
Done, feel free to edit. :-)

PR #684, opened 2024-12-06
Include notes when logging exceptions

camillol:
# Summary

This adds support for the notes feature from https://peps.python.org/pep-0678/, which was introduced in Python 3.11.
These notes are not included in `str`, but they are included in `traceback.print_exception`, etc.; we should include them when logging, too.

Note: an alternative option for formatting would be to include these in the `exc_value` string.

# Pull Request Check List


src/structlog/tracebacks.py
```diff
@@ -83,10 +83,14 @@ class SyntaxError_:  # noqa: N801
...
     """
     Represents an exception and a list of stack frames.
+
+    .. versionchanged:: 24.5.0
+       Added the *exc_notes* field.
     """
 
     exc_type: str
     exc_value: str
+    exc_notes: str | None = None
```
hynek, line 93: hey sorry for the long delay – is there a reason why you chose to make it an optional str?

You mention it in the description but without reasoning and I think it would be more ergonomic to make it a `tuple[str, ...]`?

(tuple > list because it would keep Stack hashable)

camillol: I forget; I think I was concerned about how it would show up in a particular system we have, but for no good reason. A tuple works better. Done! Sorry I lost track of this for a while.

hynek: thanks! but ugh I'm so dumb, there already _is_ a list in the class so it makes no sense to make this inconsistently a list. I'll fix it myself, this is on me.

PR #687, opened 2024-12-23
Bypass zizmor warnings

maresb:
# Summary

I suspect that your zizmor workflow is misconfigured in a way that makes it easy to bypass in an individual PR. This is an attempted proof of concept.

The fix is to use `branches: ["**"]` instead of `branches: ["*"]` since I can bypass the workflow by putting a slash in my branch name.

I recently set up zizmor myself, and I was comparing your config with the standard one, and noticed this very minor discrepancy.

# Pull Request Check List

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -20,8 +20,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-        with:
-          persist-credentials: false
       - uses: hynek/setup-cached-uv@v2
 
       - name: Run zizmor 🌈


maresb:
Ah, nevermind, sorry for the noise!

image

I will raise the converse issue in zizmor.

maresb:
I was confused about what's being pattern-matched. It's not the source branch `bugfix/no-zizmor` but the target branch `main`. So you're safe unless you create a branch in your base repo containing a slash and merge an unchecked PR into that. (Super obscure.)

Reference: official documentation for the pattern matching syntax: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

PR #709, opened 2025-03-08
TimeStamper() now uses TZ-aware objects

### LINKED ISSUE #703: Use aware time object in `processors.TimeStamper()`
At this moment `%z`, `%:z` and `%Z` time format codes are replaced by empty strings (see [Note 6](https://docs.python.org/3/library/datetime.html#technical-detail)), because of:

https://github.com/hynek/structlog/blob/768931a79f2ae08e7a42778289eec2e1b94f8c27/src/structlog/processors.py#L528-L529

Could you replace `datetime.datetime.now()` with `datetime.datetime.now().astimezone()`?

And, frankly, I disagree with the comment: I use aware time object here _exactly_ because I format it (using `%:z` in my case).

---

Off-topic: thank you for attrs and structlog! And huuuge thanks for your articles—I've learned from your texts a lot.

hynek:
The default output doesn't change but manual formatting allows for TZ data now.

Fixes #703

cc @andrei-korshikov

### DIFF
--- a/src/structlog/processors.py
+++ b/src/structlog/processors.py
@@ -541,7 +541,7 @@ def _make_stamper(
         def stamper_iso_local(event_dict: EventDict) -> EventDict:
             # We remove the timezone offset for backwards-compatibility. If the
             # user wants a timezone, they have to set fmt manually.
-            event_dict[key] = now().isoformat().rsplit("+", 1)[0]
+            event_dict[key] = now().isoformat()[:-6]
             return event_dict
 
         def stamper_iso_utc(event_dict: EventDict) -> EventDict:
--- a/src/structlog/processors.py
+++ b/src/structlog/processors.py
@@ -525,8 +525,7 @@ def _make_stamper(
     else:
 
         def now() -> datetime.datetime:
-            # A naive local datetime is fine here, because we only format it.
-            return datetime.datetime.now()  # noqa: DTZ005
+            return datetime.datetime.now().astimezone()
 
     if fmt is None:
 
@@ -540,7 +539,9 @@ def _make_stamper(
     if fmt.upper() == "ISO":
 
         def stamper_iso_local(event_dict: EventDict) -> EventDict:
-            event_dict[key] = now().isoformat()
+            # We remove the timezone offset for backwards-compatibility. If the
+            # user wants a timezone, they have to set fmt manually.
+            event_dict[key] = now().isoformat().rsplit("+", 1)[0]
             return event_dict
 
         def stamper_iso_utc(event_dict: EventDict) -> EventDict:
--- a/src/structlog/processors.py
+++ b/src/structlog/processors.py
@@ -525,7 +525,9 @@ def _make_stamper(
     else:
 
         def now() -> datetime.datetime:
-            return datetime.datetime.now().astimezone()
+            # We don't need the TZ for our own formatting. We add it only for
+            # user-defined formats later.
+            return datetime.datetime.now()  # noqa: DTZ005
 
     if fmt is None:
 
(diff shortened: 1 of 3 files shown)


src/structlog/processors.py
```diff
@@ -540,7 +539,9 @@ def stamper_unix(event_dict: EventDict) -> EventDict:
     if fmt.upper() == "ISO":
 
         def stamper_iso_local(event_dict: EventDict) -> EventDict:
-            event_dict[key] = now().isoformat()
+            # We remove the timezone offset for backwards-compatibility. If the
+            # user wants a timezone, they have to set fmt manually.
+            event_dict[key] = now().isoformat().rsplit("+", 1)[0]
```
andrei-korshikov, line 544: Counterexample:
```python
>>> datetime.datetime.now().astimezone(datetime.timezone(datetime.timedelta(hours=-5))).isoformat().rsplit("+", 1)[0]
'2025-03-08T09:55:44.415778-05:00'
```

[datetime.isoformat()](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat) states that its ISO time zone designator format is `+HH:MM[:SS[.ffffff]]`, but right below they give us a sample with negative offset—`'2009-11-27T00:00:00.000100-06:39'`.

I.e. their `+` in `+HH:MM` is actually "plus or minus". Negative offsets are totally valid by ISO 8601 / RFC 3339 anyway.

hynek: bleh true. great, according to [the docstring](https://github.com/python/cpython/blob/72e5b25efb580fb1f0fdfade516be90d90822164/Lib/_pydatetime.py#L1605), the TZ is always `+zz:zz` so before we start imitating using strftime, just a `[:-6]`?

andrei-korshikov: My proposal: `datetime.datetime.now().astimezone().replace(tzinfo=None).isoformat()`,
i.e. `event_dict[key] = now().replace(tzinfo=None).isoformat()`

[datetime.astimezone()](https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone) reads: "If you merely want to remove the timezone object from an aware datetime dt without conversion of date and time data, use dt.replace(tzinfo=None)." I think it is exactly our case: we want to convert aware time object back to naive, i.e. remove `tzinfo`.

hynek: wait, this is stupid. with either approach, I'm adding and removing the timezone instead of adding it only when necessary. i.e. in `stamper_fmt`. behold the genius who finally read his own code: https://github.com/hynek/structlog/pull/709/commits/9e0e47e344bde50b2ee9130b1bc206fb1a2ba03b

andrei-korshikov: Well, [isoformat()](https://github.com/python/cpython/blob/72e5b25efb580fb1f0fdfade516be90d90822164/Lib/_pydatetime.py#L1602) uses [_tzstr()](https://github.com/python/cpython/blob/72e5b25efb580fb1f0fdfade516be90d90822164/Lib/_pydatetime.py#L1578) which in turn uses [_format_offset()](https://github.com/python/cpython/blob/72e5b25efb580fb1f0fdfade516be90d90822164/Lib/_pydatetime.py#L187), which gives us [sing](https://github.com/python/cpython/blob/72e5b25efb580fb1f0fdfade516be90d90822164/Lib/_pydatetime.py#L197) and possibly seconds and microseconds.

I can't imagine real worlds applications for TZ offset with seconds and its fractions (some crazy astronomy stuff??), but is valid per current standards, so just `[:-6]` is not enough because time zone designator has variable length.

Yes, docstrings for these functions are confusing indeed, they repeatedly use `+hh:mm` which really is just one possible option...

hynek: since we posted at almost the same time, check back my comment before. I think I solved it pretty much perfectly.

andrei-korshikov: Yes, I do agree. I also tested it right now and it works as intended. Thanks to both of us:D

PR #723, opened 2025-05-27
Add 3.14

hynek:
The only change necessary is that we can't rely on isEnabledFor anymore, because it returns always False in the way we use it.


hynek:
the failing mypy check should start passing once GitHub ships 3.14.0b2 (it already works for me locally).

PR #729, opened 2025-06-28
Handle exception cycles

hinthornw:
When extracting frames for dict tracebacks, it's possible for exception cause chains to have cycles if a user calls `raise e from <something in its chain>`. Tracking the seen IDs would let the traceback break early in this case.

Thanks for maintaining a great library!

### DIFF
--- a/src/structlog/tracebacks.py
+++ b/src/structlog/tracebacks.py
@@ -239,8 +239,13 @@ def extract(
 
     stacks: list[Stack] = []
     is_cause = False
+    seen: set[int] = set()
 
     while True:
+        if id(exc_value) in seen:
+            break
+        seen.add(id(exc_value))
+
         stack = Stack(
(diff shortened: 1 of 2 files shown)


hynek:
hey sorry for the late answer – it's been… busy. Is this a fix to the same thing as #727? It looks similar but the approach looks very different?

cc @sscherfke

hinthornw:
No, it is not the same thing as #727
No, main does not fix this.

My minimal example is in the test case I added to this pr.

Or shorter:
```
try:
   raise ValueError("...")
except Exception as e:
    raise e from e
```

Creates an infinite recursion in the logging traceback.

I know that code snippet is dumb. I have a server implementation that executes user code. It uses structlog because you all have made a great library.  User code can do somewhat dumb things like this code snippet. Structlog's current behavior makes a single error (impacting 1 request) terminal for the server.

PR #777, opened 2025-12-11
feat: EventRenamer handle event=None

bentinata:
# Summary

Let `EventRenamer` defaults to `None` in case `event` key is missing.

## Long one (more context)

I'm new to python development in general. I am adding `structlog` to my `uvicorn` app, so that the logs are parseable by the logs processor. My current orgs setup expect `message` key, hence I am using `EventRenamer`. But when my logger get passed `None` or `[]` (empty array), it will break the logger.

## Other considerations and questions:

Maybe let `None` be passed as `event`?

In the `_base.py` `_process_event`, there's this if-clause.

```py
if event is not None:
    event_dict["event"] = event
```

I wonder why do this?

## Other approach

I've added new custom processors like this:

```py
def ensure_event_key(_, __, event_dict: EventDict) -> EventDict:
    if 'event' not in event_dict:
        event_dict['event'] = ''
    return event_dict
```

# Pull Request Check List

I haven't done anything more than just this small change. I would like to validate my understanding first.

### DIFF
--- a/src/structlog/processors.py
+++ b/src/structlog/processors.py
@@ -960,7 +960,7 @@ class EventRenamer:
     def __call__(
         self, logger: logging.Logger, name: str, event_dict: EventDict
     ) -> EventDict:
-        event = event_dict.pop("event")
+        event = event_dict.pop("event", None)
         event_dict[self.to] = event
 
         if self.replace_by is not None:


hynek:
Hi Ben, sorry for not getting back to you earlier.

> # Summary
> Let `EventRenamer` defaults to `None` in case `event` key is missing.
> 
> ## Long one (more context)
> I'm new to python development in general. I am adding `structlog` to my `uvicorn` app, so that the logs are parseable by the logs processor. My current orgs setup expect `message` key, hence I am using `EventRenamer`. But when my logger get passed `None` or `[]` (empty array), it will break the logger.

What kind of configuration are you using exactly? IIUC, you're violating the typing contract of the log methods that all require event to be a string?

I'm somewhat sure that you're doing something you shouldn't based on some misunderstanding so it would be better to sort that out first.
 
> ## Other considerations and questions:
> Maybe let `None` be passed as `event`?
> 
> In the `_base.py` `_process_event`, there's this if-clause.
> 
> ```python
> if event is not None:
>     event_dict["event"] = event
> ```
> 
> I wonder why do this?

I *think* this is from the olden days with a more dynamic logger implementation where `event` was in fact optional. Although I'm looking at stdlib and it looks like the type hints there are still `str | None`. 🙈

bentinata:
> sorry for not getting back to you earlier.

Please don't be sorry. Open source work is hard. I didn't demand your reply. Thanks for this, though.

> you're violating the typing contract of the log methods that all require event to be a string?

You could say that. I am working with some internal library that may returns `None`. I'm asking your opinion as a more experienced Python developer: Would wrapping those logger calls with `f'{}'` in my code be better? Wouldn't it look "noisy"? E.g.:

```py
# x may be None
logger.info(f'{x}')
```

> Although I'm looking at stdlib and it looks like the type hints there are still str | None.

Did you mean these?

https://github.com/hynek/structlog/blob/212a69497e4f4eb93b94a324eaabf258ba1cf70e/src/structlog/stdlib.py#L218

Would strengthening the type here helps? Or would it cause more problem?

Thanks.

hynek:
> > sorry for not getting back to you earlier.
> 
> Please don't be sorry. Open source work is hard. I didn't demand your reply. Thanks for this, though.

❤️

> > you're violating the typing contract of the log methods that all require event to be a string?
> 
> You could say that. I am working with some internal library that may returns `None`. I'm asking your opinion as a more experienced Python developer: Would wrapping those logger calls with `f'{}'` in my code be better? Wouldn't it look "noisy"? E.g.:
> 
> ```python
> # x may be None
> logger.info(f'{x}')
> ```

This looks weird because what exactly is an event "None"? I would go for something like `log.info("appropriate_event", result=x)`

`appropriate_event` can be something like "computed_whatever" or "called_some_api".

>  > Although I'm looking at stdlib and it looks like the type hints there are still str | None.
> 
> Did you mean these?
> 
> https://github.com/hynek/structlog/blob/212a69497e4f4eb93b94a324eaabf258ba1cf70e/src/structlog/stdlib.py#L218
> 
> Would strengthening the type here helps? Or would it cause more problem?

The problem here is that (and I just checked), standard library logging allows for None as a log message:

```
>>> import logging
>>> logging.getLogger().critical(None)
None
```

Not sure what to do here without breaking stuff, but your problem seems to lie elsewhere.

Issue #715, opened 2025-03-31
dict_traceback showing locals which includes credentials


leoskyrocker:
Sensitive data can be printed by exposing all of the locals.
I've encountered one with using psycopg2. It stores database connection strings, and is exposed when there is an uncaught exception.

#### Proposed Resolution
Allow dict traceback or traceback.SHOW_LOCALS to be configured (currently it is hardcoded)

sscherfke:
Are you using `structlog.processors.dict_tracebacks()`?

This is just a shortcut for `ExceptionRenderer(ExceptionDictTransformer())` which has a `show_locals` argument.

If secrets are wrapped with `pydantic.Secret()` or `typed_settings.Secret()`, they will not leak.  But I think very few people actually to this and for external libs, you can't control this behavior.

So maybe changing the default of `SHOW_LOCALS` to `False` might be a good idea.

hynek:
fixed by #817

PR #782, opened 2026-01-14
added sensitive data redactor processor

yk0007:
needed a way to redact sensitive fields before logs hit storage. handles passwords, api keys, etc.

supports exact field names and glob patterns like *password*. also handles nested dicts.

related to #715


hynek:
I'm sorry but this is way too big and complex to land in structlog proper. I would suggest to publish it on PyPI and link it from [here](https://github.com/hynek/structlog/wiki/Third-party-Extensions).

PR #788, opened 2026-02-20
Fix double-repr in to_repr for truncated non-string objects

bysiber:
`to_repr()` in the fallback (non-Rich) path applies `!r` to `obj_repr[:max_string]`, but `obj_repr` is already the result of `repr(obj)`. The `!r` format specifier calls `repr()` again on the already-repr'd slice, wrapping it in spurious quotes and escaping characters.

**Before (truncated list with max_string=10):**
```
'[1, 2, 3, '+5
```

**After:**
```
[1, 2, 3, +5
```

The `str`/`bytes` branch on the lines above correctly applies `!r` to the raw `obj[:max_string]` (not yet repr'd), so it doesn't have this problem. The fix simply drops the `!r` since `obj_repr` is already a string from `repr()`.


hynek:
this is not a bug. the `"'foo'+x"` indicates it's a fragment.

PR #794, opened 2026-03-14
Monochrome Rich traceback rendering w/ ConsoleRenderer(colors=False)

funkyfuture:
# Summary

As addressed in [this SO question](https://stackoverflow.com/questions/79905552/structlogs-consolerenderer-writes-tracebacks-with-color-control-codes-to-file-t), exceptions are logged with colours when using the `ConsoleRenderer` with the default rich exception formatter even when the renderer was initialized w/ `colors=False`.

This patch selects monochrome output for that formatter when it's used as default.

# Pull Request Check List


hynek:
> thanks for the hint! improperly fiddling w/ global objects was the cause.

it usually is ;)

thanks, i've fixed it up. fun, surprising, fact: `x in (y, z)` counterintuitively compares x to y and z by equality, not identity.

PR #806, opened 2026-04-13
Expose remove_file_from_lock helper for WRITE_LOCKS cleanup

wjddn279:
# Summary

Hi, I'm an active user and contributor of Airflow. While investigating a memory leak in Airflow, I discovered that structlog's WRITE_LOCKS holds references to file descriptors created by the process, and these references prevent the associated loggers from being garbage collected even after they are no longer in use. I was able to confirm that the memory leak is resolved by importing WRITE_LOCKS and manually removing the references.

please reference the pr related: https://github.com/apache/airflow/pull/65121

As far as I can tell, there is currently no public interface for removing file references from WRITE_LOCKS. I don't think it's ideal for users to directly import and manipulate a variable from a private module. I've added a public interface that allows users to clean up WRITE_LOCKS references — I'd appreciate it if you could review it.

# Pull Request Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,12 @@ You can find our backwards-compatibility policy [here](https://github.com/hynek/
 
 ## [Unreleased](https://github.com/hynek/structlog/compare/25.5.0...HEAD)
 
+### Added
+
+- `structlog.remove_file_from_lock()` to release the write lock that is implicitly registered for a file by `PrintLogger`, `WriteLogger`, and `BytesLogger`.
+  Useful when the underlying file is being closed.
+
+
 ### Removed
 
 - Python 3.8 and 3.9 support.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -64,6 +64,8 @@ API Reference
 
 .. autoclass:: BytesLoggerFactory
 
+.. autofunction:: remove_file_from_lock
+
 .. autoexception:: DropEvent
 
 .. autoclass:: BoundLoggerBase
--- a/src/structlog/__init__.py
+++ b/src/structlog/__init__.py
@@ -37,6 +37,7 @@ from structlog._output import (
     PrintLoggerFactory,
     WriteLogger,
     WriteLoggerFactory,
+    remove_file_from_lock,
 )
 from structlog.exceptions import DropEvent
 from structlog.testing import ReturnLogger, ReturnLoggerFactory
@@ -79,6 +80,7 @@ __all__ = [
     "is_configured",
     "make_filtering_bound_logger",
     "processors",
+    "remove_file_from_lock",
     "reset_defaults",
     "stdlib",
     "testing",
--- a/src/structlog/_output.py
+++ b/src/structlog/_output.py
@@ -30,8 +30,17 @@ def _get_lock_for_file(file: IO[Any]) -> threading.Lock:
     return lock
 
 
-def remove_file_from_lock(file: IO[Any]):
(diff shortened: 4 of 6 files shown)


ashb:
Not yet sure if we need this change on the structlog side.

hynek:
OK lmk one y'all sure. :)

wjddn279:
@hynek 

Hi! I have a question regarding this :)

Is there a specific intention behind not having a separate release logic for WRITE_LOCKS? I would expect the reference to be released when the structlog object is properly closed or cleaned up, but with the current behavior, it seems like there is no way to remove the reference from WRITE_LOCKS.

hynek:
Could this be not solved with weak references? Sounds to me like they were made for such use-cases:

> A weak reference to an object is not enough to keep the object alive: when the only remaining references to a referent are weak references, [garbage collection](dfile:///Users/hynek/Library/Application%20Support/Dash/DocSets/Python/Python.docset/Contents/Resources/Documents/doc/glossary.html#term-garbage-collection) is free to destroy the referent and reuse its memory for something else. However, until the object is actually destroyed the weak reference may return the object even if there are no strong references to it.

wjddn279:
I tested replacing it with a weak reference — I tried passing log_file_descriptor as a weakref instead of passing it directly.
```
log_file_descriptor = log_file.open("ab")
proxy = weakref.proxy(log_file_descriptor)
underlying_logger = structlog.BytesLogger(proxy)
logger = structlog.wrap_logger(underlying_logger, logger_name="task").bind()
```
```
  File "/Users/user/Dev/airflow/airflow-core/src/a.py", line 29, in <module>
    logger, log_file_descriptor = _configure_logging(log_path)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/Dev/airflow/airflow-core/src/a.py", line 14, in _configure_logging
    underlying_logger = structlog.BytesLogger(proxy)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/Dev/airflow/.venv/lib/python3.12/site-packages/structlog/_output.py", line 266, in __init__
    self._lock = _get_lock_for_file(self._file)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/Dev/airflow/.venv/lib/python3.12/site-packages/structlog/_output.py", line 25, in _get_lock_for_file
    lock = WRITE_LOCKS.get(file)
           ^^^^^^^^^^^^^^^^^^^^^
TypeError: unhashable type: 'weakref.ProxyType'
```
It ended up throwing an error when accessing WRITE_LOCKS. Is this the intended behavior you had in mind?

wjddn279:
@hynek 

Hi, I wanted to check whether you see the missing WRITE_LOCK release as a bug that should be fixed. If that's the case, I'd be happy to submit a PR for it. I'm raising this because I have some concerns about how it's currently handled on the consumer side.

hynek:
> Hi, I wanted to check whether you see the missing WRITE_LOCK release as a bug that should be fixed. If that's the case, I'd be happy to submit a PR for it. I'm raising this because I have some concerns about how it's currently handled on the consumer side.

I hope #807 fixes your problem, because that seems to be the correct way to fix this?

wjddn279:
yes. LGTM thanks! @ashb

hynek:
great! when I suggested wekarefs to you, I missed that the _keys_ were the referenced part and wanted to check back once I have time to breath, thankfully @ashb took that off my plate 🥳

PR #812, opened 2026-05-10
Add CallsiteParameter.QUAL_MODULE

### LINKED ISSUE #492: Adding ability to print full namespace of module for CallsiteParameterAdder

hynek:
Fixes #492

### DIFF
--- a/src/structlog/processors.py
+++ b/src/structlog/processors.py
@@ -785,7 +785,7 @@ def _get_callsite_qual_name(module: str, frame: FrameType) -> Any:
 
 
 def _get_callsite_qual_module(module: str, frame: FrameType) -> Any:
-    return frame.f_globals.get("__name__", "__main__")
+    return module
 
 
 def _get_callsite_lineno(module: str, frame: FrameType) -> Any:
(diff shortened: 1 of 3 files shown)


src/structlog/processors.py
```diff
@@ -779,6 +784,10 @@ def _get_callsite_qual_name(module: str, frame: FrameType) -> Any:
     return frame.f_code.co_qualname  # will crash on Python <3.11
 
 
+def _get_callsite_qual_module(module: str, frame: FrameType) -> Any:
+    return frame.f_globals.get("__name__", "__main__")
```
decaz: @hynek why not just `return module`?

hynek: ahhh you're right `_find_first_app_frame_and_name` calls f_globals.get like 5 times to get module 🙈

PR #817, opened 2026-05-28
Add a recipe for customizing exc dict formatting

sirosen:
A new recipe titled "Custom Filtering of Dict Tracebacks" shows how to
subclass an ExceptionDictTransformer, dig around in the chain of
exceptions, sift through "locals", and make in-place modifications.

As a good demo of the sort of expensive function you might want to tie
to such operations, the example defines "high entropy strings" as
anything with a Shannon Entropy > 4.0 (which is arbitrary but somewhat
low).

In the `exceptions` documentation, a link is provided to this example
along with a direct indication that users should start by looking at the
parameters for `ExceptionDictTransformer`, which covers most cases.

resolves #789, #715

---

The example given is what sprang to mind for me, but I'm happy to rewrite it in any way that seems suitable.
(e.g., we could replace entropy with some simple bogus check like `len(data) == 42`)

# Summary

Add a new Recipe doc which shows how you can deeply customize dict tracebacks.

# Pull Request Check List

_NB: many check-list items marked as "vacuously true", as this is a pure docs change._

### DIFF
--- a/docs/exceptions.md
+++ b/docs/exceptions.md
@@ -26,10 +26,12 @@ The most common use-cases are already covered by the following processors:
 
 : Uses {class}`structlog.tracebacks.ExceptionDictTransformer` to give you a structured and JSON-serializable `exception` key.
 
+
 ### Customizing Dict Tracebacks
 
 {class}`structlog.tracebacks.ExceptionDictTransformer` can be customized with a number of parameters.
 If these prove insufficient, see {ref}`Custom Filtering of Dict Tracebacks <custom-dict-tracebacks>`.
+
 
 ## Console rendering
 
--- a/docs/exceptions.md
+++ b/docs/exceptions.md
@@ -26,6 +26,10 @@ The most common use-cases are already covered by the following processors:
 
 : Uses {class}`structlog.tracebacks.ExceptionDictTransformer` to give you a structured and JSON-serializable `exception` key.
 
+### Customizing Dict Tracebacks
+
+{class}`structlog.tracebacks.ExceptionDictTransformer` can be customized with a number of parameters.
+If these prove insufficient, see {ref}`Custom Filtering of Dict Tracebacks <custom-dict-tracebacks>`.
 
 ## Console rendering
 
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -141,6 +141,62 @@ These two methods and one attribute are all you need to write own *bound loggers
 [dry]: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself
 
 
+(custom-dict-tracebacks)=
+
+## Custom Filtering of Dict Tracebacks
+
+Dict traceback printing via {class}`structlog.tracebacks.ExceptionDictTransformer` may expose values in logs which you do not want.
+For example, tokens and credentials are easily captured in locals and logged.
+
+As a simple control, you can set `show_locals=False`.
+For more extensive customization, subclass the transformer.
+
+{class}`ExceptionDictTransformer <structlog.tracebacks.ExceptionDictTransformer>`'s `__call__` is responsible for converting exception info into a list of dicts.
+Because this is only called during exception handling, it's an ideal point of control for potentially expensive filtering functionality.
+
+The following example customizes the transformer with a measure of string entropy, and redacts values in `locals` which exceed a threshold.
+
+```
+import math
+
+from structlog.tracebacks import ExceptionDictTransformer
+
+
+def is_high_entropy_string(data):
+    if not isinstance(data, str):
+        return False
+    charset = set(data)
+    entropy = 0
+    for c in charset:
+        p_x = data.count(c) / len(data)
+        entropy -= p_x * math.log2(p_x)
+    return entropy > 4.0
+
+
+def redact_locals(frame_locals):
+    redact_values = set()
+    for k, v in frame_locals.items():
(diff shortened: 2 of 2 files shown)


hynek:
hmm, I find the entropy idea cute, but it wouldn't even trigger on a credit card number, so it feels a bit risky to put that into docs and people might start relying on it since it looks sciencey and magical.

I think it would be more conservative/safer to look for keys like `token`, `password` etc? I wouldn't mind keeping it as a fallback I guess.

sirosen:
Yeah, that danger seems very real. I'd like to think folks are more sensible but it has logarithms and therefore is clearly maths magic! :grin: 
Why don't we just nix it and look for `token` and `password`, nothing else? I'll put together a new revision.

I fished for an easy-to-explain expensive computation, but the risk of someone putting the example thoughtlessly into production scares me off of it.

hynek:
Yeah Twisted has a class called [InMemoryUsernamePasswordDatabaseDontUse](https://docs.twisted.org/en/stable/api/twisted.cred.checkers.InMemoryUsernamePasswordDatabaseDontUse.html) and AFAICT it’s the most-used creds checker in the ecosystem. Seeing a log call is mythical attraction to normies. ;)

sirosen:
Oh, I'm not sure if the doc should mention this, but I did make a small example script to test.
<details><summary><code>custom_exception_rendering.py</code></summary>

```python
import structlog
import structlog.processors

from structlog.tracebacks import ExceptionDictTransformer


def redact_locals(frame_locals):
    return {
        k: (v if k not in ("token", "password") else "<REDACTED>")
        for k, v in frame_locals.items()
    }


def redact_frames(exc_dict):
    for frame in exc_dict["frames"]:
        if "locals" in frame:
            frame["locals"] = redact_locals(frame["locals"])
    return exc_dict


class RedactingExceptionDictTransformer(ExceptionDictTransformer):
    def __call__(self, exc_info):
        exceptions = super().__call__(exc_info)
        return [redact_frames(exc) for exc in exceptions]


logger = structlog.getLogger(__name__)


structlog.configure(
    processors=[
        structlog.processors.ExceptionRenderer(
            RedactingExceptionDictTransformer()
        ),
    ]
)


def func():
    token = "abcdefg0123456789"
    try:
        {}[token]
    except Exception:
        logger.exception("oh no")


func()
```
</details>

One of the nice things that running the demo shows is that the string is redacted in `locals` but not in exception text. Maybe a note on this should go after the example in the recipe? Basically "don't turn off your brain, locals aren't the only thing that matters."

hynek:
> Oh, I'm not sure if the doc should mention this, but I did make a small example script to test.
> 
> `custom_exception_rendering.py`

This is something frustrating that I should fix but will not block this release on. My most modern project actually does this well: https://github.com/hynek/svcs/tree/main/docs/examples

`docs` has a whole `examples` subdirectory where the examples are included into the docs and that allows me to have tests next to them.

> One of the nice things that running the demo shows is that the string is redacted in `locals` but not in exception text. Maybe a note on this should go after the example in the recipe? Basically "don't turn off your brain, locals aren't the only thing that matters."

Sure, maybe mention that concrete thing, too?

sirosen:
I added a little note after the example. LMK if you want to wordsmith it a bit more, but I think it's good.

---

> This is something frustrating that I should fix but will not block this release on.

I don't want to make you block on me, but I'd love to help port the pattern from `svcs` back to here.

It would let me see how you've done things, and I've wanted to do this for years on tools like `webargs` but never worked out the details.

If that appeals, let me know and I'll file an issue and try to lay it out sensibly.

docs/recipes.md
```diff
@@ -141,6 +141,48 @@ These two methods and one attribute are all you need to write own *bound loggers
...
+
+class RedactingExceptionDictTransformer(ExceptionDictTransformer):
+    def __call__(self, exc_info):
+        exceptions = super().__call__(exc_info)
+        return [redact_frames(exc) for exc in exceptions]
+```
+
+Please remember that values may appear elsewhere in the dict traceback, for example in an exception string.
+If you remove data from `locals`, that does not guarantee that it is removed entirely from the log.
+
```
hynek, line 185:
```suggestion


```

docs/exceptions.md
```diff
@@ -26,6 +26,10 @@ The most common use-cases are already covered by the following processors:
 
 : Uses {class}`structlog.tracebacks.ExceptionDictTransformer` to give you a structured and JSON-serializable `exception` key.
 
+### Customizing Dict Tracebacks
+
+{class}`structlog.tracebacks.ExceptionDictTransformer` can be customized with a number of parameters.
+If these prove insufficient, see {ref}`Custom Filtering of Dict Tracebacks <custom-dict-tracebacks>`.
```
hynek, line 33:
```suggestion
If these prove insufficient, see {ref}`Custom Filtering of Dict Tracebacks <custom-dict-tracebacks>`.

```

docs/exceptions.md
```diff
@@ -26,6 +26,10 @@ The most common use-cases are already covered by the following processors:
 
 : Uses {class}`structlog.tracebacks.ExceptionDictTransformer` to give you a structured and JSON-serializable `exception` key.
 
+### Customizing Dict Tracebacks
```
hynek, line 30:
```suggestion

### Customizing Dict Tracebacks
```

hynek (approved):
Thanks, this should be good enough for now!

PR #828, opened 2026-07-20
100% type coverage

jorenham:
# Summary

Following #825, this fills in the last couple of missing annotations, bringing the `structlog` public type coverage to a beautiful 100% :tada:

Before:

```console
$ pyrefly coverage check --public-only src/structlog
 WARN `structlog.contextvars.STRUCTLOG_KEY_PREFIX_LEN` is untyped [coverage-missing]
  --> src/structlog/contextvars.py:35:1
   |
35 | STRUCTLOG_KEY_PREFIX_LEN = len(STRUCTLOG_KEY_PREFIX)
   | ------------------------
   |
 WARN `structlog.dev.rich_traceback` is untyped [coverage-missing]
   --> src/structlog/dev.py:471:5
    |
471 | /     def rich_traceback(*args, **kw):
472 | |         raise ModuleNotFoundError(
473 | |             "RichTracebackFormatter requires Rich to be installed.",
474 | |             name="rich",
475 | |         )
    | |_________-
    |
 WARN `structlog.processors.format_exc_info` is untyped [coverage-missing]
   --> src/structlog/processors.py:427:1
    |
427 | format_exc_info = ExceptionRenderer()
    | ---------------
    |
 WARN `structlog.processors.dict_tracebacks` is untyped [coverage-missing]
   --> src/structlog/processors.py:448:1
    |
448 | dict_tracebacks = ExceptionRenderer(ExceptionDictTransformer())
    | ---------------
    |
 WARN `structlog.threadlocal.ThreadLocal` is untyped [coverage-missing]
  --> src/structlog/threadlocal.py:47:1
   |
47 | ThreadLocal = _determine_threadlocal()
   | -----------
   |
ERROR type coverage 98.89% (622 of 629 typable) is below the 100.00% threshold
```

After:

```console
$ pyrefly coverage check --public-only src/structlog
 INFO type coverage 100.00% (630 of 630 typable)
```

# Pull Request Checklist

- [x] I acknowledge this project's [**AI policy**](https://github.com/hynek/structlog/blob/main/.github/AI_POLICY.md).
- [x] This pull request is [**not** from my `main` branch](https://hynek.me/articles/pull-requests-branch/).
  - Consider granting [push permissions to the PR branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), so maintainers can fix minor issues themselves without pestering you.
- [ ] There's **tests** for all new and changed code.
[...]

PR #831, opened 2026-07-27
Enforce 100% type coverage in CI with Pyrefly

jorenham:
# Summary

Following https://github.com/hynek/structlog/pull/828#issuecomment-5078389391, this ensures that the public API type coverage is kept at 100% by running `pyrefly coverage check` in CI.

There should be no need to route this through tox, because `pyrefly` already supports `uvx` natively. 

# Pull Request Checklist

- [x] This pull request is [**not** from my `main` branch](https://hynek.me/articles/pull-requests-branch/).
- [ ] There's **tests** for all new and changed code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,9 +148,6 @@ jobs:
           uvx --with tox-uv
           tox run -f typing
 
-      - name: Enforce 100% public type coverage with Pyrefly
-        run: uvx pyrefly coverage check --public-only src/structlog
-
   docs:
     name: Run doctests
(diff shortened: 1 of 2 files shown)


hynek (changes requested):
> There should be no need to route this through tox, because pyrefly already supports uvx natively.

That is true, but I try to have everything that needs to pass offline too with online/offline parity, such that pushing a passing tox has good chances to pass in CI too.

Just add it to the existing pyrefly tox env and call it a day – thanks!

hynek:
sweet, thanks!

PR #835, opened 2026-08-01
Introduce register_log_level/reset_log_levels

sirosen:
# Summary

_This is open as a draft for initial feedback. I want to confirm whether or not the overall approach is acceptable before working on fleshing out documentation and refining the work further._

This PR is a potential solution for #47.

These two methods, housed in a dedicated module, call out to register custom log levels across various other components. Registering a log level does the following:

- confirm the level and name are good
- add the level and name to mappings
- add a relevant named method to structlog-provided loggers
- add the log level to level-based filtering bound loggers

The reset method inverts these steps, removing from these same registered locations in reverse order.

Names which are already used by `structlog` are forbidden, so no special checking or "undo rules" are needed for the reset.

# Pull Request Checklist

- [x] This pull request is [**not** from my `main` branch](https://hynek.me/articles/pull-requests-branch/).


src/structlog/_custom_log_levels.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    while CUSTOM_LOG_LEVEL_NAMES:
+        name = CUSTOM_LOG_LEVEL_NAMES.pop()
+        remove_custom_level_filtering(name)
+        remove_custom_level_method(name)
+        del NAME_TO_LEVEL[name]
+
+    while CUSTOM_LOG_LEVELS:
+        level = CUSTOM_LOG_LEVELS.pop()
+        del LEVEL_TO_NAME[level]
```
sirosen, line 112: Is a new module the right place for all of this? `stdlib` didn't feel right to me.

src/structlog/_custom_log_levels.py
```diff
@@ -0,0 +1,112 @@
...
+        "warn",
+        "awarn",
+        "warning",
+        "awarning",
+        "msg",
+        "amsg",
+        # NOTSET is a special 0-valued level, but not a method name
+        "notset",
+    )
+)
```
sirosen, line 48: If these names are to be forbidden, I think statically listing them is more legible than trying to be fancy/dynamic. But there's a risk of drift...
Plus, forbidding the names means that adding a new method to this list could be a breaking change! Yuck!

This is one of the things I winced at.

src/structlog/_native.py
```diff
@@ -246,7 +248,7 @@ async def alog(
 
     meths: dict[str, Callable[..., Any]] = {"log": log, "alog": alog}
     for lvl, name in LEVEL_TO_NAME.items():
-        meths[name], meths[f"a{name}"] = make_method(lvl)
+        meths[name], meths[f"a{name}"] = _make_method(min_level, lvl)
```
sirosen, line 251: The diff is illegible, but all I did is pop out `make_method`, add `min_level` to its args, and dedent it. I needed that for the usage below.

src/structlog/_native.py
```diff
@@ -284,3 +286,24 @@ async def alog(
...
+
+    LEVEL_TO_FILTERING_LOGGER[level] = _make_filtering_bound_logger(level)
+
+
+def remove_custom_level_filtering(level_name: str, /) -> None:
+    level = NAME_TO_LEVEL[level_name]
+    for bound_logger in LEVEL_TO_FILTERING_LOGGER.values():
+        delattr(bound_logger, level_name)
+        delattr(bound_logger, f"a{level_name}")
+    del LEVEL_TO_FILTERING_LOGGER[level]
```
sirosen, line 309: Alternatively, these methods could fully rebuild the `LEVEL_TO_FILTERING_LOGGER` table each time? But that's a lot of repeat work each time a log level gets registered.

src/structlog/_output.py
```diff
@@ -374,3 +374,15 @@ def __init__(self, file: BinaryIO | None = None):
...
+def add_new_custom_level_method(level_name: str, /) -> None:
+    """Register a new log level method."""
+    for logger_class in (BytesLogger, PrintLogger, WriteLogger):
+        setattr(logger_class, level_name, logger_class.msg)
+
+
+def remove_custom_level_method(level_name: str, /) -> None:
+    """Remove all registered log level methods."""
+    for logger_class in (BytesLogger, PrintLogger, WriteLogger):
+        delattr(logger_class, level_name)
```
sirosen, line 388: I thought it's slightly more manageable if some of the responsibility here is passed to the component modules. You can see a similar bit of logic added to `_native`, but this is more clearly isolated.

As above though, we have to patch with setattr/delattr and there's a real feeling here that it would be very easy to make a mistake. e.g., Add a new logger class but not add it here.

sirosen (commented):
I've done a self-review to note my high-level questions and concerns.

